PR #102, opened 2013-01-03
Add delete-by stream primitive and test case

lwf:
delete-by allows you to delete events from the index based on properties of the event. We use it to delete individual metrics (delete-by index [:host :service]) when we're performing upgrades but also to delete all metrics for a particular host when we retire hosts (delete-by index :host).


vipinvkmenon:
has this been removed?

PR #347, opened 2014-03-30
Implement an index backend for HSQLDB

eric:
## Overview

This pull request provides a new index backend that uses an embedded in-memory [HSQLDB](http://hsqldb.org/) to store the events.

## The Problem

The current `index` implementation requires scanning of the entire index for all queries. This can become CPU intensive as the size of the index and the number of queries per second grows.

## The Solution

To speed up queries, we convert the queries to SQL and use indexes on all of the queryable columns.

## The State Of Things

This pull request is still a work in progress and is **not ready to be merged**.

The majority of the code has been written, but I'm sure there's a lot more to do. It hasn't been run with any clients, so I'm sure there will be more issues that that will hash out.

No load testing of any kind has been done yet, so how this performs in the real world compared to the existing index is not currently known.

## Remaining Items

- [x] Implement `delete-exactly`
- [x] Deal with Outstanding Questions
- [ ] Add documentation to all functions


aphyr:
Note to self: definitely review and merge this.

pyr:
hi @eric, are you still willing to finish the documentation bits? this seems like a valid alternative for high-cardinality, low write and heavy queried indices.

eric:
That's an interesting idea. We would definitely need to test it in those situations to see how it does.

Because it didn't end up helping my situation, it isn't something that I've ended up using. I think it's interesting from an illustrative perspective, but I'm unsure if it would help someone in production or not.

pyr:
@eric thanks for the quick feedback, tabling for now then.

PR #423, opened 2014-07-28
allowing the 'ws-server' to process plain http GET requests

tempredirect:
This is a port of the changes proposed in https://github.com/aphyr/riemann/pull/422

I've switched the uri's to /index and the 'q' parameter is now 'query' to be consistent with the existing websockets uris.

provides:
    GET /index list of all events in the index
                query parameters:
                'offset' - offset of first item to return
                'limit'  - max number to return (defualt is 250)
                'query'  - riemann query to match events against

### DIFF
--- a/src/riemann/common.clj
+++ b/src/riemann/common.clj
@@ -117,16 +117,16 @@
         [[:time (unix-time)]
          [:state "expired"]]))
 
-(defn event-with-iso8601-time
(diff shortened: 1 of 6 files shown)


aphyr:
I don't think this applies to the rewritten websockets/sse servers any more; gonna close. :-/

PR #436, opened 2014-08-29
Add plugin for output to Keen IO

gphat:
Like it says on the tin! :smile: 

We're clojure noobs and this was largely snagged from the Pagerduty and OpenTSDB plugins.

We like to batch events for performance reasons but I didn't see any examples of outputs that batch things up. I know this can be accomplished inside riemann but it will still cause us to send multiple POSTs instead of one big one periodically. Thoughts?

Thanks for any tips, improvements or suggestions.

### DIFF
--- a/src/riemann/config.clj
+++ b/src/riemann/config.clj
@@ -29,6 +29,7 @@
         [riemann.plugin  :only [load-plugin load-plugins]]
         [riemann.time :only [unix-time linear-time once! every!]]
         [riemann.pagerduty :only [pagerduty]]
+        [riemann.keenio :only [keenio]]
         [riemann.campfire :only [campfire]]
         [riemann.kairosdb :only [kairosdb]]
         [riemann.librato :only [librato-metrics]]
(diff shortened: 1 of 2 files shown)


aphyr:
I wrote a macro for mocking IO to make this a little easier. Check out https://github.com/aphyr/riemann/blob/master/test/riemann/shinken_test.clj#L12 for an example of how to verify you're making the right calls to clj-http. :)

aphyr:
OK, I wrote a basic test for this and merged it; GH doesn't see the merge but I promise it's there. Thanks @gphat! :)

gphat:
Oh, shit man. That's nice of you. I'm sorry I'm such a slacker. We got asked by a client if this integration was possible so I jumped at the chance to give it a go then lol stuff happened and it fell off my radar and ya know.

Thanks ever so much for a) keeping this on your mind and b) fixing our shortcomings.

:sparkles:  :heart:  :sparkles:

PR #456, opened 2014-10-15
hand over logging configuration to logconfig

pyr:
This first step just swaps `logging/init` with `start-logging!` from logconfig


pyr:
will open a new one to leverage the updated [unilog](https://github.com/pyr/unilog) project.

PR #471, opened 2014-12-13
A simplistic query limit implementation

algernon:
I'm not entirely sure whether this is the best solution. Especially the `riemann.index/search` change feels wrong. But it shows signs of working, so I'm submitting it for consideration, even if there's quite a bit of work left to do.

### DIFF
--- a/src/riemann/Query.g
+++ b/src/riemann/Query.g
@@ -18,6 +18,7 @@ tokens {
 	GREATER = '>';
 	GREATER_EQUAL = '>=';
 	TAGGED = 'tagged';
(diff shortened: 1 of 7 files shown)


src/riemann/Query.g
```diff
@@ -48,6 +49,7 @@ simple	:	( t | f | nil
 		| greater_equal
 		| not_equal
 		| equal
+		| limit
```
algernon: I'll play with putting it further up. Thanks for the suggestion!

aphyr:
Gonna close this for now.

PR #481, opened 2015-01-01
Added example of stable to down service detection

jamtur01:


aphyr:
Well said. Thanks James. :)

PR #485, opened 2015-01-06
Include original event when capturing call rescue exception

tcrayford:
adds the original event to the event passed to `call-rescue` as a custom attribute, so there's more debugging context

### DIFF
--- a/src/riemann/common.clj
+++ b/src/riemann/common.clj
@@ -145,7 +145,7 @@
                 :service "riemann exception"
                 :state "error"
                 :tags ["exception" (.getName (class e))]
-                :original-event original
+                :event original
                 :description (str e "\n\n"
                                   (join "\n" (.getStackTrace e)))})))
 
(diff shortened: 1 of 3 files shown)


src/riemann/common.clj
```diff
@@ -139,13 +139,15 @@
...
-               :tags ["exception" (.getName (class e))]
-               :description (str e "\n\n"
-                                 (join "\n" (.getStackTrace e)))}))
+  ([exception] (exception->event exception nil))
+  ([^Throwable e original]
+   (map->Event {:time (unix-time)
+                :service "riemann exception"
+                :state "error"
+                :tags ["exception" (.getName (class e))]
+                :original-event original
```
aphyr: Wanna call it :event?

tcrayford:
new change now: include the original exception in the event sent to downstream as well.

This means downstream folk that wanna do things with e.g. the stacktrace don't have to parse it out of the event.

tcrayford:
oh, and re the test failure: https://twitter.com/t_crayford/status/553965338971635712

aphyr:
Cool, the maven boundary repo disappeared? That's gonna be fun.

aphyr:
Looks good to me! Thanks Tom. :)

PR #487, opened 2015-01-08
Fix Netty native transport issue

dbriones:
Use `NioServerSocketChannel` and `NioEventLoopGroup` instead of the corresponding `Epoll` classes, since the latter are Linux-specific.

### DIFF
--- a/src/riemann/transport/tcp.clj
+++ b/src/riemann/transport/tcp.clj
@@ -72,7 +72,7 @@
 
     (isSharable [] true)))
 
-(def transport
+(def netty-implementation
   (if (.contains (. System getProperty "os.name") "Linux")
     {:event-loop-group-fn #(EpollEventLoopGroup.)
      :channel EpollServerSocketChannel}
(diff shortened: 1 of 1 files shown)


aphyr:
Aside: you can build packages via `lein pkg`. Might help you get a fix in prod faster.

I'd like to do platform detection here and use EPoll where possible, because it's offers decent speedup in my benchmarks.

dbriones:
Thanks! Yeah, I'm going to build my fork and install it in prod. 

I'm happy to look into platform detection, as well...  I think I might create a branch off of this one and do a separate PR for that.

aphyr:
Should be pretty straightforward! Here's an example: https://github.com/aphyr/riemann/blob/9548d5801ff75a6d7251c88a5c9a5fa4f621bcbe/src/riemann/bin.clj#L37

dbriones:
OK, back to you, @aphyr. I didn't test this particularly rigorously -- basically just verified that the TCP server would start up on OS X. You should probably do the same on Linux :)

dbriones:
Yay! Build passed. As you can see, I did all the work on this PR since the changes weren't that large and I couldn't be bothered to create a new branch :)

Back to you, @aphyr!

aphyr:
Terrific, thanks @dbriones! I'll push a release today.

PR #491, opened 2015-01-12
Fix init script output on RedHat-based distros

pharaujo:
Starting Riemann on CentOS/RHEL would output no feedback whether it started successfully or otherwise. I'm assuming the `success` and `failure` functions are defined because the script already assumes `/etc/rc.d/init.d/functions` is available for source'ing.


aphyr:
Sounds good to me. I don't have a redhat distro handy so I'll defer to your tests here. :)

PR #492, opened 2015-01-12
restore one netty metric

sgran:
It's a start.  I'm still not sure how to get at QueueSize (if there even is such a concept now).

### DIFF
--- a/src/riemann/transport.clj
+++ b/src/riemann/transport.clj
@@ -144,13 +144,10 @@
 
     (reify Instrumented
       (events [this]
-        (comment (let [base {:state "ok" :time (unix-time)}]
+        (let [base {:state "ok" :time (unix-time)}]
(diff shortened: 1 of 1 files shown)


aphyr:
Thank you, @sgran! I did some digging into MultiThreadEventExecutor but I don't see a queue... anywhere, actually. I'm wondering the handler queues have been pushed to a higher level somewhere. I'll ask Maurer about it. :)

aphyr:
Digging a bit more, looks like we can call (iterator-seq (.iterator shared-event-executor)) to get a seq of active eventexecutors, and when those are instances of http://netty.io/4.0/api/io/netty/util/concurrent/SingleThreadEventExecutor.html (which hopefully they are), we can call .pendingTasks() on each to figure out how many it's got queued. Sum those puppies up to get an overall metric.

PR #493, opened 2015-01-12
Evaluate where* predicate expression once

### LINKED ISSUE #381: where* evaluates predicate expression every time
We should bind where*'s predicate expression using let, then call that bound value as a function with each event.

AeroNotix:
Closes #381


aphyr:
Ah shoot, I actually *just* committed this, haha. :)

AeroNotix:
I has a sad

PR #494, opened 2015-01-12
Emit a deprecation warning only once

### LINKED ISSUE #362: deprecation notices should only print once
We don't want to spam the log unnecessarily

AeroNotix:
Closes #362

### DIFF
--- a/src/riemann/common.clj
+++ b/src/riemann/common.clj
@@ -22,15 +22,16 @@
   (match [pred object]
     "Does predicate describe object?"))
 
-; Deprecation
+(def deprecations-emitted (atom {}))
+
 (defmacro deprecated
   "Wraps body in an implicit (do), and logs a deprecation notice when invoked."
   [comment & body]
-  `(do
-     (info ~(str "Deprecated: "
-                 (format "<%s:%s> " *file* (:line (meta &form)))
-                 comment))
-     ~@body))
+  (let [id (str *file* ":" (:line (meta &form)) " - " comment)]
+    (swap! deprecations-emitted assoc id (delay (info comment)))
+    `(do
+       (force (get @deprecations-emitted ~id))
+       ~@body)))
 
 (def hostname-refresh-interval
   "How often to allow shelling out to hostname (1), in seconds."


src/riemann/common.clj
```diff
@@ -32,6 +32,17 @@
...
+  [name comment ds args & body]
+  `(let [emit-depr-warning# (delay
+                              (info ~(str "Deprecated: "
+                                       (format "<%s:%s> " *file* (:line (meta &form)))
+                                       comment)))]
+     (defn ~name ~args ~ds
+       (do
+         (force emit-depr-warning#)
+         ~@body))))
+
```
aphyr: Could we make this behavior part of (deprecated) instead of defining a second macro?

AeroNotix: I actually wanted to remove `deprecated` altogether.

Do you want to keep both? There were no instances where a body of code needed to be deprecated over a whole function.

aphyr:
What I'd really like to do, BTW, is set up the logger to say things like "last message repeated 1500 times", because there's more than just deprecation notices at play here. Would be nice for IO failures and so on where the log gets spammed with the same notice over and over.

AeroNotix:
@aphyr That's a bit different from logging deprecations. Do you want to track that in a separate ticket?

src/riemann/common.clj
aphyr: I figured this would come in handy when we start deprecating specific behaviors of functions, like "X can no longer take a single arity", etc.

aphyr:
Yeah, I'll create a separate ticket for it. I'm fine with this as a stopgap.

AeroNotix:
@aphyr :+1:

src/riemann/common.clj
AeroNotix: I was trying to implement it in terms of that, but my macro-fu couldn't achieve the macrology necessary to do that. Any tips?

aphyr: Right, cuz the expanded form gets invoked multiple times. Umm...

```clj
(def deprecations-emitted (atom {}))

(defmacro deprecated [...]
  ; Make up an ID for this warning
  (let [id (str FILE ":" LINE " - " msg )]
    ; Set up a global delay for this form.
    (swap! deprecations-emitted assoc id (delay (info ...))))
  ; In generated code, look up and force realization of that delay
  `(do (force (get @deprecations-emitted id)) ~@body))
```

AeroNotix: Yupp, was trying to avoid the global. I'll make the changes, thanks!

AeroNotix:
@aphyr I made the requested changes. Sorry it took so long, life etc.

aphyr:
Thanks, Aaron. :)

PR #496, opened 2015-01-16
Add streams/fixed-wall-clock-time-window

juise:


aphyr:
Hi @juise! Thanks for taking the time to write this.

I'm... having a bit of trouble understanding the motivation behind this function, and how we can clearly explain the different semantics to users. Can you elaborate a bit on the use cases?

juise:
Hi!

Right now I'm use riemann as accumulator and aggregator in dozens of workers, that push every 5 or 30 seconds aggregated metrics into main riemann that writes aggregated metrics into influxdb. For aggregation from dozens of workers with big metric value modulation (by example from 1k to 400 in 3 seconds) with big time windows like 5 or 30 seconds very necessary to know what latency you have.

By example: if the server 'b' is started within 3 seconds after the server 'a', the aggregated metric via fixed-time-window with 5 secs and corresponding to same interval of time may fall into different aggregated events, but with fixed-time-wall-clock-window metrics corresponding to same interval of time will fall into same aggregated event.

Here - https://www.dropbox.com/s/puaci5hahxdnm4u/ftw.jpg I try to explain it.

aphyr:
Ahhh, I see where you're going here: you want every stream to choose the same window alignment. I got confused because one of them invokes unix-time instead of relying on the event times. Can I suggest just doing the window computation as (mod n) on the event times instead? Relying on the stream time instead of unix-time means we can window events correctly even when they're delayed relative to wall-clock time.

juise:
You are right, I think it will be more correct use (mod (:time event) n) instead of unix-time, fixed! And one more thing, may be better name will be smth like fixed-alignment-time-window?

aphyr:
Yeah, I think fixed-alignment-time-window would be fine. I think I use `offset` to refer to this kind of window alignment in riemann.time, so perhaps `fixed-offset-time-window` would be more consistent?

juise:
I think it's important to keep names consistency in project, so lets it will be fixed-offset-time-window!

juise:
great, thanks!

PR #497, opened 2015-01-16
Fix deb and rpm ownerships

mfournier:
This squashes the ownership discrepencies between deb and rpm packages and fixes an issue reported on the mailing-list.

This would be the resulting ownserships if this patchset is merged:

owned by riemann:
 - `/etc/riemann`
 - `/var/log/riemann`

owned by root:
 - `/usr/bin/riemann`
 - `/usr/{share,lib}/riemann`
 - `/etc/init.d/riemann`
 - `/etc/{default,sysconfig}/riemann`


aphyr:
Yeah, this looks like the right call. Thanks for taking the time to clean this up, @mfournier!

PR #499, opened 2015-01-21
Update clj-http

AeroNotix:
Ref: https://github.com/ztellman/aleph/issues/134

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -21,7 +21,7 @@
     [net.logstash.log4j/jsonevent-layout "1.5"]
     [com.cemerick/pomegranate "0.3.0"]
     [org.spootnik/http-kit "2.1.18.1"]
-    [clj-http "0.9.1"]
+    [clj-http "1.0.1"]
     [cheshire "5.3.1"]
     [clj-librato "0.0.5"]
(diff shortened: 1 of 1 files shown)


AeroNotix:
This works for me locally, any ideas?

aphyr:
I don't think we should explicitly depend on Potemkin; it's a clj-http dependency; makes sense to let them handle it.

aphyr:
Build errors look like a clj-http/potemkin version mismatch. Bet you if you drop the potemkin dep it'll work out OK.

AeroNotix:
@aphyr indeed. The context is in that ticket I linked. I guess I'll wait till clj-http makes another release before using aleph with Riemann.

aphyr:
Done and done, thanks for the bump Aaron! :)

AeroNotix:
:+1:

aphyr:
So whatever bug this is seems related to clj-http 1.0.1; I don't see any conflicting deps for potemkin, so I'm not sure why 1.0.1 should have problems. Dropping to clj-http 1.0.0 works correctly. cc @ztellman

PR #500, opened 2015-01-26
Implemented Boundary integration.

blkt:
Boundary's integration should be similar in use to mailer in that it does not produce immediately a function that acts on streams, but a function that accepts additional arguments to simplify some use cases. See documentation for details.

Please, let me know also if tests are acceptable.

### DIFF
--- a/src/riemann/boundary.clj
+++ b/src/riemann/boundary.clj
@@ -25,50 +25,45 @@
 
   Examples:
 
-  ((boundarify) \"foo\") => \"FOO\"
-  ((boundarify) \"foo bar\") => \"FOO_BAR\"
-  ((boundarify) \"foo@\") => \"FOO\"
-  ((boundarify) \"foo@bar\") => \"FOOBAR\"
-  ((boundarify \"org\") \"foo\") => \"ORG_FOO\"
-  ((boundarify) \"!#@\") => exception
-  ((boundarify \"org\") \"!#@\") => exception
+  (boundarify \"foo\") => \"FOO\"
+  (boundarify \"foo bar\") => \"FOO_BAR\"
+  (boundarify \"foo@\") => \"FOO\"
+  (boundarify \"foo@bar\") => \"FOOBAR\"
+  (boundarify \"foo\" \"org\") => \"ORG_FOO\"
+  (boundarify \"!#@\") => exception
+  (boundarify \"!#@\" \"org\") => exception
   "
-  [& [organization]]
-  (fn [service]
-    (let [good-ones (s/replace (s/upper-case (s/replace service #"\s+" "_"))
-                               #"[^A-Z0-9_]" "")
-          res (if-not (nil? organization)
-                (str (s/upper-case organization) "_" good-ones)
-                good-ones)]
-      (if-not (empty? good-ones)
-        res
-        (throw (RuntimeException.
-                (str "can't accept the given service string \""
-                     service "\" as metric id")))))))
+  [service & [organization]]
+  (let [good-ones (-> service
+                      (s/replace #"\s+" "_")
+                      s/upper-case
+                      (s/replace #"[^A-Z0-9_]" ""))]
+    (when (empty? good-ones)
+      (throw (RuntimeException.
+              (str "can't accept the given service string \""
+                   service "\" as metric id"))))
+    (if (nil? organization)
+      good-ones
+      (str (s/upper-case organization) "_" good-ones))))
 
 (defn ^:private packer-upper
-  "Returns a function that packs up the events in a form suitable for
+  "Returns a function packs up the events in a form suitable for
   Boundary's API.
 
   If a metric-id is given, it will be used for all the events in the
   pack. Otherwise, every single event service is \"boundarified\". In
   both cases, organization is prepended if given."
-  [metric-id organization]
+  [{:keys [metric-id org]}]
   (let [helper
         #(vector
           (:host %)
-          `~(cond (and (nil? metric-id) (nil? organization))
-                  ((boundarify) (:service %))
-                  (and (nil? metric-id) (not (nil? organization)))
-                  ((boundarify organization) (:service %))
-                  (and (not (nil? metric-id)) (nil? organization))
-                  ((boundarify) metric-id)
-                  :else
-                  ((boundarify organization) metric-id))
+          (if (nil? metric-id)
+            (boundarify (:service %) org)
+            (boundarify metric-id org))
           (:metric %)
           (:time %))]
     (fn [events]
-      (vec (map helper events)))))
+      (mapv helper events))))
 
 (defn boundary
   "Returns a function used to generate specific senders (like mailer)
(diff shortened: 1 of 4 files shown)


src/riemann/boundary.clj
```diff
@@ -0,0 +1,98 @@
...
+  ((boundarify) \"foo\") => \"FOO\"
+  ((boundarify) \"foo bar\") => \"FOO_BAR\"
+  ((boundarify) \"foo@\") => \"FOO\"
+  ((boundarify) \"foo@bar\") => \"FOOBAR\"
+  ((boundarify \"org\") \"foo\") => \"ORG_FOO\"
+  ((boundarify) \"!#@\") => exception
+  ((boundarify \"org\") \"!#@\") => exception
+  "
+  [& [organization]]
+  (fn [service]
```
aphyr: Why curry this function? It doesn't look like there are any circumstances where it's not immediately invoked then discarded.

src/riemann/boundary.clj
```diff
@@ -0,0 +1,98 @@
...
+  Boundary's API.
+
+  If a metric-id is given, it will be used for all the events in the
+  pack. Otherwise, every single event service is \"boundarified\". In
+  both cases, organization is prepended if given."
+  [metric-id organization]
+  (let [helper
+        #(vector
+          (:host %)
+          `~(cond (and (nil? metric-id) (nil? organization))
```
aphyr: Why quote and immediately unquote here?

src/riemann/boundary.clj
```diff
@@ -0,0 +1,98 @@
...
+    (let [good-ones (s/replace (s/upper-case (s/replace service #"\s+" "_"))
+                               #"[^A-Z0-9_]" "")
+          res (if-not (nil? organization)
+                (str (s/upper-case organization) "_" good-ones)
+                good-ones)]
+      (if-not (empty? good-ones)
+        res
+        (throw (RuntimeException.
+                (str "can't accept the given service string \""
+                     service "\" as metric id")))))))
```
aphyr: Perhaps clearer like this?

```clj
    (let [good-ones (-> service
                        (s/replace #"\s+" "_")
                        s/upper-case
                        (s/replace #"[^A-Z0-9_]" ""))]
      (when (empty? good-ones)
        (throw (IllegalArgumentException.
                 (str "can't accept the given service string "
                      (pr-str service) " as metric id"))))
      (if (nil? organization)
        good-ones
        (str (s/upper-case organization) "_" good-ones)))
```

src/riemann/boundary.clj
```diff
@@ -0,0 +1,98 @@
...
+  construct the destination Boundary's metric id. In both cases,
+  organization is prepended if non nil.
+
+  Examples:
+
+  (def bdry (boundary eml tkn))
+  (when :foo (bdry)) => builds the destination metric id with :service
+  (when :foo (bdry :metric-id \"METRIC_ID\")) => sends to METRIC_ID"
+  [email token]
+  (fn [& {:keys [metric-id org] :or {metric-id nil org nil}}]
```
aphyr: For Riemann argument handling, we try to use maps instead of destructuring args; you can just drop the `&` here.

Why provide nil defaults? Isn't nil the default for maps anyway?

src/riemann/boundary.clj
```diff
@@ -0,0 +1,98 @@
...
+                  (and (nil? metric-id) (not (nil? organization)))
+                  ((boundarify organization) (:service %))
+                  (and (not (nil? metric-id)) (nil? organization))
+                  ((boundarify) metric-id)
+                  :else
+                  ((boundarify organization) metric-id))
+          (:metric %)
+          (:time %))]
+    (fn [events]
+      (vec (map helper events)))))
```
aphyr: `(mapv f coll)` is probably faster/clearer than `(vec (map f coll))`.

src/riemann/boundary.clj
```diff
@@ -0,0 +1,98 @@
...
+  If a metric-id is given, it will be used for all the events in the
+  pack. Otherwise, every single event service is \"boundarified\". In
+  both cases, organization is prepended if given."
+  [metric-id organization]
+  (let [helper
+        #(vector
+          (:host %)
+          `~(cond (and (nil? metric-id) (nil? organization))
+                  ((boundarify) (:service %))
+                  (and (nil? metric-id) (not (nil? organization)))
```
aphyr: (not (nil? organization)) is implied by the previous branch; I think you can drop it here.

src/riemann/boundary.clj
```diff
@@ -0,0 +1,98 @@
...
+                  ((boundarify organization) (:service %))
+                  (and (not (nil? metric-id)) (nil? organization))
```
aphyr: Ditto here: the first branch precludes both being `nil`, so we can just check for `(nil? organization)`

src/riemann/boundary.clj
blkt: This is a legacy of some attempts I made to generate exactly the right access pattern depending on whether you're going to use a specific `metric-id` or `:service` field.

I forgot about it and it is useless. I'll fix it.

src/riemann/boundary.clj
blkt: Sure.

src/riemann/boundary.clj
```diff
@@ -0,0 +1,98 @@
...
+                  ((boundarify) metric-id)
+                  :else
+                  ((boundarify organization) metric-id))
```
aphyr: Since organization defaults to `nil` in `(boundarify)`, I'm not sure all four branches are actually necessary here.

test/riemann/boundary_test.clj
```diff
@@ -0,0 +1,56 @@
...
+
+    (testing "boundary (no authentication)"
+      (is (function? (boundary "eml" "tkn")))
+      (is (function? ((boundary "eml" "tkn"))))
+      (is (function? ((boundary "eml" "tkn") :metric-id "metric")))
+      (is (function? ((boundary "eml" "tkn") :organization "org")))
+      (is (function? ((boundary "eml" "tkn")
+                      :metric-id "metric" :organization "org")))
+      (is (function? ((boundary "eml" "tkn")
+                      :organization "org" :metric-id "metric"))))))
```
aphyr: See https://github.com/aphyr/riemann/blob/master/test/riemann/keenio_test.clj#L8-L26 for an example of an end-to-end test that mocks out the HTTP call.

test/riemann/boundary_test.clj
```diff
@@ -0,0 +1,56 @@
+(ns riemann.boundary-test
+  (:use riemann.boundary
+        [riemann.time :only [unix-time]]
+        clojure.test)
+  (:require [riemann.logging :as logging]))
+
+(logging/init)
+
+(deftest ^:boundary ^:integration boundary-test
```
aphyr: Since this test doesn't do any network IO, there's no need to tag it as `:integration`.

test/riemann/boundary_test.clj
blkt: Thanks, I was looking for something similar around Riemann sources.

test/riemann/boundary_test.clj
blkt: Fixing it.

test/riemann/boundary_test.clj
```diff
@@ -0,0 +1,56 @@
...
+  (let [b @#'riemann.boundary/boundarify
+        p @#'riemann.boundary/packer-upper
+        time (unix-time)]
+
+    (testing "boundarify correctness"
```
aphyr: This is a big enough test that you might consider breaking it up at function boundaries: `(deftest boundarify-test ...)` etc.

src/riemann/boundary.clj
blkt: Does this policy apply to all functions, even to private, utility ones (like packer-upper)?

aphyr: Yeah, I try to maintain it as a standard. Makes function composition and partial defaults easier, and performs faster.

blkt: Ok, I'll keep that in mind.

aphyr:
Still got some formatting/spelling issues, but I'll take it. Thanks for all your hard work here, @blkt. :)

blkt:
Glad I could contribute :D

PR #501, opened 2015-01-26
cache maven downloads between builds

tcrayford:
this caches maven downloads between builds, which requires the use of
travis ci's new caching infrastructure. Should speed up builds quite a
bit


aphyr:
Cool, thanks @tcrayford

PR #502, opened 2015-01-27
Removed daemonize

jamtur01:
The daemonize package hasn't been updated for recent EPEL releases,
specifically 7 onwards. That means this RPM won't install on RHEL or
CentOS 7 or later.

I've replaced daemonize with daemon to get around this.

### DIFF
--- a/pkg/rpm/init.sh
+++ b/pkg/rpm/init.sh
@@ -32,8 +32,9 @@ DAEMON_OPTS="${RIEMANN_OPTS} ${RIEMANN_CONFIG}"
 start() {
     echo -n $"Starting ${NAME}: "
     ulimit -n $NFILES
-#    daemon --pidfile $PID_FILE --user $RIEMANN_USER $DAEMON $DAEMON_OPTS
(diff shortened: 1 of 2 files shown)


aphyr:
Works for me. You've tested this, I presume? Is this gonna break Centos6?

mfournier:
+1 as long as this doesn't break on older versions of centos.

FWIW, I opened a packaging request for daemonize on epel7 a couple of months ago: https://bugzilla.redhat.com/show_bug.cgi?id=1144474

Annoying I admit, but porting this package from el6 is straightforward. Just grab the .src.rpm and run `rpmbuild -bb`.

jamtur01:
@mfournier I'm having some issues getting `daemon` to behave so it's still a testing exercise. I agree porting the package is easy but I'd love to be able to use a CM tool and EPEL rather than rebuild an RPM and ship that around. :(

jamtur01:
I'm stumped. daemon is crap. :( Closing this as a result. If folks can push on EPEL that'd be great.

PR #507, opened 2015-02-04
Apply configuration during tests

AeroNotix:
I'm not sure if this is the right way to go about it but my use-case is that I want to be able to query the index during tests.

Perhaps we could use the same method which is used for `riemann.tests/*streams*`?

### DIFF
--- a/src/riemann/bin.clj
+++ b/src/riemann/bin.clj
@@ -76,6 +76,7 @@
                 (reset! config-file config)
                 (riemann.config/include @config-file)
                 (binding [test/*streams* (:streams @config/next-core)]
+                  (riemann.config/apply!)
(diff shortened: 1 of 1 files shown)


AeroNotix:
Could someone restart the travis build, looks unrelated to my changes.

aphyr:
The reason why we don't apply the config is because we don't want to try to start network servers that would conflict with the running Riemann instance.

AeroNotix:
How would I go about using index queries for tests? What about a similar approach to `riemann.tests/*streams*`?

aphyr:
With what we have now, I think your best bet is just to wrap the index with a tap. Getting the full config infrastructure to run in a pure way is a bit more of a stretch.

PR #512, opened 2015-02-12
Fix :ttl conversion to LIFETIME in xymon forward

mirwan:
:ttl is in seconds whereas xymon status message's LIFETIME parameter is expected in minutes


aphyr:
Cool, thanks @mirwan! :)

PR #516, opened 2015-02-17
Retain time specified in event

fhalim:
The timestamp from the event should be retained in the influxdb point.


aphyr:
I was *wondering* why they didn't send the time too! Good to know influx will accept a time field. :)

PR #520, opened 2015-02-26
Put EXTRA_CLASSPATH after riemann.jar

md5:
Ensures that Riemann's copy of it's classes and Clojure get loaded first.

This prevents the issue where an uberjar in `EXTRA_CLASSPATH` that includes an earlier version of Clojure can interfere with Riemann's classloading.

### DIFF
--- a/pkg/deb/riemann
+++ b/pkg/deb/riemann
@@ -4,7 +4,7 @@ if [ -f /etc/default/riemann ]; then
     . /etc/default/riemann
 fi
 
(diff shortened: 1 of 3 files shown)


aphyr:
Works for me! Thanks @md5. :)

md5:
:metal:

md5:
Do you have any plans to release a new version any time soon? It looks like there are a decent number of changes on `master` that haven't been released yet: https://github.com/aphyr/riemann/compare/0.2.8...c8c42ff

aphyr:
Yeah, just blocking on the query parser/compiler rewrite.

PR #525, opened 2015-03-13
expire by branch if event is expired

itaifrenkel:
This commit removes by branch streams if an expired event flows through them. 
This allows using the by stream even if the resulting number of streams would be big - as long as an expire event is triggered to remove the old streams.

### DIFF
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -1507,7 +1507,10 @@
              fork (if-let [fork (@table fork-name)]
                     fork
                     ((swap! table assoc fork-name (new-fork)) fork-name))]
-         (call-rescue event fork)))))
+         (do
+           (if (expired? event) (swap! table dissoc fork-name))
+           (call-rescue event fork))))))
+
 
 (defn changed
   "Passes on events only when (f event) differs from that of the previous
--- a/test/riemann/streams_test.clj
+++ b/test/riemann/streams_test.clj
@@ -772,6 +772,34 @@
            (s {:metric 1})
            (is (= @i 2))))
 
+(deftest by-expire-starts-new-branch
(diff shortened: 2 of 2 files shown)


itaifrenkel:
@aphyr I am concerned about this change not being backwards compatible. What's your take on this?

aphyr:
This is the behavior I intended for (by), but yeah... reasoning about expiry is hard. The case that comes to mind is detecting and collapsing *runs* of expired events, but my feeling is that GC is *necessary* and expiry is the right place to do it; asking folks to transform expiry into a state like :critical if they're doing flap-suppression seems OK to me.

aphyr:
So yeah, with all that said, I vote we do it, and I'll write a changelog entry warning folks of the semantics change. I think it'll help more than it hurts, personally. :)

aphyr:
Ohhhh hang on, on second thought there might be an issue here. `by` doesn't necessarily partition based on [:host :service] every time. We need to keep track of the currently valid [:host :service] tuples which have flowed into each fork, and make sure that list is *empty* before releasing the reference to that fork. Otherwise, in (by :host) the first service to expire on a host will nuke *every* service on that host. :)

itaifrenkel:
If someone does (by [:host]) the result would be unpredictable even with the provisioning your proposed due to weird timings and event/expiry race conditions due to the periodic nature of expired events. 

I propose two more constrained alternatives:
(by-host-service ) same behavior as (by [:host :service] ) just with expiry
(by-bye ...) same behavior as by just with expiry (any expired event) documenting the common use case (by-bye [:host :service]) and the caveats in any other usage.

itaifrenkel:
continued here https://github.com/aphyr/riemann/pull/526

PR #526, opened 2015-03-14
revert expiration support of by. the new by-host-service supports expiration

itaifrenkel:


aphyr:
Ehhh, I dunno if I really want to add a second function for this--folks are gonna be confused about which one to use and the implications. I'd rather find a way to have `by` do the right state machine all the time.

itaifrenkel:
ok. so would the following logic satisfy that? 
If fields == [:host :service] or fields == [:service :host] 
then expire-branches? = true 
else expire-branches? = false

The reason I'm suggesting this simplistic approach is not to add memory consumption on the (by) clause by performing elaborate book-keeping.

aphyr:
So right now nobody's using (by) with that many streams--I think the memory cost should be relatively small. The big question is how expensive is the machinery for figuring out whether a fork should expire, and I think for that I'd like to see benchmarks demonstrating the cost is significant before we add user-facing complexity. The smaller the API we expose, the better. :)

That said, if the cost *is* significant, I vote we have two functions--one which does what `by` does presently, and another which does the appropriate expiry tracking regardless of the key function--if it were just a performance optimization I'd be fine having argument-specific behavior, but these are fundamentally different behaviors and I want to make that as explicit as possible.

itaifrenkel:
I suppose the common use case you have in mind is monitoring web servers, and doing bookkeeping so when a server is dead, we can release its stream. However the bookeeping would be beneficial when you have hundreds of thousands of "dead" servers or more ( I mean how much memory does an empty stream take)? did anyone complain about that memory leak ?

The use case I have in mind is monitoring web clients! I have a backup plan that queries the index foreach incoming event. However, I do think that a specific (by-host-service) is worth checking as the resulting riemann configuration would be more straight forward than explicit index querying. As for tracking, the use case for [:host :service] does not require tracking (as you can see from the pull request). 

What use cases do you have in mind that requires tracking ?

aphyr:
So far folks have been fine using Riemann with small cardinalities for `by`, but eventually we'd like to support broad time-varying cardinalities, especially for memory-intensive streams.

itaifrenkel:
What I am saying is that the use case you are referring to has not materialized yet. The use case I am referring to is very real to me and does not require any bookkeeping. I suspect once people hear that you can monitor web clients with riemann it would not be as isoteric as it sounds now, and then it would not require bookkeeping either. Could you please provide concrete use cases that would require bookkeeping?

aphyr:
The reason you submitted this patch in the first place was because you needed time-varying high-cardinality `by` streams, right?

itaifrenkel:
I need high-cardinality (by [:host :service]) streams. yes.

AeroNotix:
I've used riemann with 20k+ streams on `by`. Not a noticeable memory overhead. AFAICR the overhead per stream is the cost of an IFn instance.

aphyr:
Memory cost might be really cheap, might be really expensive, depending on what data the stream stores. There are some scheduler costs too. Might be fine for itai's use case, but I'm always up for doing it more efficiently!

itaifrenkel:
Let's say each web client TTL is 1 minute, and let's take a modest 100K concurrent connections. Then after 1 hour I would have 6M streams, after one day I would have 144M streams. After one week I would get to 1G streams that needs garbage collection.

aphyr:
So I spent four solid days trying to make this work, and in the process exposed some hopelessly awful things about the Riemann streaming model, expiration, etc. It involved literally hundreds of lines of code, dozens of interacting branches, all kinds of contention issues... fucking nightmare. I think (by) GC requires a fundamental redesign of the streaming model to be done correctly, and I'm not willing to inflict the horror I created on Riemann users just yet. Just too many unpredictable changes to semantics. :(

itaifrenkel:
@aphyr We're running with (by-host-service) instead of (by [:host :service]) successfully for a few months now. 

1. All memory leaks are gone now and we are monitoring all of our javascript snippets running on browsers worldwide in realtime using riemann. In terms of time-to-market this is the wholly grail of javascript snippet development.
2. There is coupling between the fields used to index the event for TTL purposes and the (by) clause. So we have transformation functions that convert custom event data into [:host :service] before indexing and transform it back after the (by) clause. So what we actually need is a way to say which fields should be used for indexing/TTLing and then (by-ttl whichever fields it was indexed in the first place. This is a different riemann requirement than we started with, and should not be as difficult to implement.
3. Some memory leaks were due to our fault, we forgot to index some of the events, so they never expired, or indexed them in the wrong way. Again... (2) could have solved alot of it.

Great shout-out to @Tomer1510 for this incredible work, and for @MosheZada that is going to have a slide or two about it in the upcoming StatsCraft conference.

Itai

PR #529, opened 2015-03-22
Documentation updates

jamtur01:
1. Added Slack example.
2. Fixed div closure for integration section.
3. Re-formatted some integrations into sixcol format.


aphyr:
Wooo!

PR #539, opened 2015-03-26
Include config files in alphabetic order

moonranger:
By default `include` loads config files in no specific order, since `java.io.File.listFiles` (thus `file-seq`) does not guarantee the order of returned files. Sometimes we may want to guarantee it loads files in a specific order, e.g. putting common functions in a `00-core.clj` and name other files like `01-service-alerts.clj` etc so that `00-core.clj` is always loaded first to make the common functions available in other config files.

Since `java.io.File` implements `Comparable`, we can simply call `sort` after `file-seq`.

### DIFF
--- a/src/riemann/config.clj
+++ b/src/riemann/config.clj
@@ -398,7 +398,7 @@
     (binding [*config-file* path
               *ns* (find-ns 'riemann.config)]
       (if (.isDirectory file)
-        (doseq [f (file-seq file)]
+        (doseq [f (sort (file-seq file))]
           (when (config-file? f)
             (load-file (.toString f))))
         (load-file path)))))


aphyr:
We've talked about this before, and my feeling is that folks should use namespaces instead of relying on the order of side effects spread across multiple files. http://riemann.io/howto.html#organizing-with-namespaces

moonranger:
That makes sense and I agree that for more complex configuration it's a better idea to organise the config files with namespaces. But IMHO it doesn't hurt to also make the loading order consistent so users have another choice for those not-so-complex configurations.

aphyr:
Ehhhhhh I reallllly want to avoid folks ever getting into this sort of situation; the less I can encourage it the better.

moonranger:
Fair enough...

PR #540, opened 2015-03-30
Faxmodem so rcvbuf

faxm0dem:
This adds an option to set the socket receive buffer size.
We're having dramatic improvements with respect to InErrors used in conjunction with collectd-write_riemann.
Thanks @algernon for the clojure-foo


aphyr:
Cool! You should also adjust the ServiceEquiv protocol implementation for UdpServer to compare this option, so Riemann knows when config changes should result in new UDP servers. :)

faxm0dem:
Done for `:so-rcvbuf` and also for `:max-size`

aphyr:
Terrific, thanks!

PR #541, opened 2015-04-04
instrument the number of unique events in the index

tcrayford:
implements `Instrumented` for the index, with just a single event for now - the number of unique service/host keyed events in there

### DIFF
--- a/src/riemann/index.clj
+++ b/src/riemann/index.clj
@@ -98,7 +98,7 @@
       (events [this]
         (let [base {:state "ok" :time (unix-time)}]
           (map (partial merge base)
-               [{:service "riemann index event-count"
+               [{:service "riemann index size"
                  :metric (.size hm)}])))
 
       clojure.lang.Seqable
--- a/src/riemann/index.clj
+++ b/src/riemann/index.clj
@@ -10,7 +10,8 @@
   Index: indexing and querying events
   Seqable: returning a list of events
   Service: lifecycle management"
-  (:require [riemann.query :as query])
+  (:require [riemann.query :as query]
+            [riemann.instrumentation :refer [Instrumented]])
   (:use [riemann.time :only [unix-time]]
          riemann.service)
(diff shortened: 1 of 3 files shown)


tcrayford:
Query: do I have to do anything to "register" this as an instrumented Service, so that instrumentation events will get injected into streams, or is that all taken care of already?

aphyr:
https://github.com/aphyr/riemann/blob/master/src/riemann/core.clj#L21 Looks like it should be picked up automatically.

src/riemann/index.clj
```diff
@@ -93,6 +94,13 @@
       (lookup [this host service]
         (.get hm [host service]))
 
+      Instrumented
+      (events [this]
+        (let [base {:state "ok" :time (unix-time)}]
+          (map (partial merge base)
+               [{:service "riemann index event-count"
```
aphyr: For consistency with async-queue metrics, wanna call it "riemann index size"?

tcrayford:
done. Ready to merge :D

wickedshimmy:
Change the metric name in the test too :)

Big high five for adding this btw!

test/riemann/index_test.clj
```diff
@@ -71,6 +72,13 @@
            (is (= 5 (:metric (lookup i 1 1))))
            (is (= 7 (:metric (lookup i 1 2))))))
 
+(deftest nbhm-instrumentation
+  (let [i (wrap-index (index))]
+
+    (i {:host 1 :service 1 :metric 5 :time 0})
+    (i {:host 1 :service 2 :metric 7 :time 0})
+
+    (is (= 2 (:metric (first (filter #(= (:service %) "riemann index event-count") (events i))))))))
```
aphyr: Just waiting on changing "event-count" to "size" and we'll be good to go. :D

tcrayford:
ready

aphyr:
Woo!

PR #543, opened 2015-04-07
check emails recipients are strings

tcrayford:
This helps with the case where you accidentally pass an event to the
result of `mailer`, in which there won't be any errors thrown (before
this commit).


aphyr:
Ah, usability improvements. My favorite. You rock, Tom. :)

PR #544, opened 2015-04-09
Fix typo in doc-example for async-queue

joelapatatechaude:


aphyr:
Thanks, Christophe! :)

PR #546, opened 2015-04-12
Added first draft of Clojure article

jamtur01:
I added a nav link for it but y'all may want to tweak that.

Might need a pass over it to make sure it's sane too.


aphyr:
I like it! Thank you very much, James. :)

PR #549, opened 2015-04-17
Add Pushover

amir:
Even though this one is an obvious copy/paste of other forwarders, but I can vouch for its functionality.


aphyr:
Terrific! Thanks @amir :)

PR #550, opened 2015-04-18
Add -v option for riemann to display the version.

joelapatatechaude:
A need this option for a check in my nose tests.

I didn't find a way to replace riemann file inside target/rpm workarea, so I created a temp file inside target, and deleted it once the rpm is build.

I tested the build for tar, rpm and deb, in all cases the packages contains a riemann that has -v option and that displays the correct version (0.2.10-SNAPSHOT)

### DIFF
--- a/pkg/deb/riemann
+++ b/pkg/deb/riemann
@@ -19,12 +19,20 @@ Runs Riemann with the given configuration file.
 OPTIONS:
   -h    Show this message
   -a    Adds some default aggressive, nonportable JVM optimization flags.
+  -v    Output Riemann version and exit.
 
 COMMANDS:
   start    Start the Riemann server (this is the default)
   test     Run the configuration tests
 
   Any unrecognized options (e.g. -XX:+UseParNewGC) will be passed on to java.
+EOF
+}
+
+version()
+{
+  cat <<EOF
+_VERSION_
 EOF
 }
 
@@ -36,6 +44,10 @@ for arg in "$@"; do
       ;;
     "-h")
       usage
+      exit 0
+      ;;
+    "-v")
+      version
       exit 0
       ;;
     -*)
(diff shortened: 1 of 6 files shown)


aphyr:
I'm a little confused about how this works--do we install this VERSION file to every node? What path does it wind up under?

joelapatatechaude:
What is the 'VERSION file' and what do you mean by 'node'?

1. Both rpm, deb and tar package directory has their own riemann script, and the riemann script has existing -a and -h options. So I thought this would be the natural place to add the -v option. Was this a correct assumptions?

2. I need somehow to get the version information from the riemann script. I decided to add a placeholder (_VERSION_) in the (three versions of) riemann and insert the version at build time. Again, is it a correct way to proceed?

3. I now need a way to update the _VERSION_ ) at build time.
a. for fattar, instead of COPY the riemann file from pkg/tar/riemann to bin/riemann, I use the spit / slurp consruct to be able to replace _VERSION_ on the fly.
b. Similar for the fatdeb
c. For the fatrpm it is a little bit more involved, but essentially the same thing.

I am happy to work more on this, but I would need some instructions on how this should be done as I have no experience with building clojure code.

bfritz:
@joelapatatechaude Rather than write a new file, you could parse `pom.properties` archived in the uberjar, e.g.

    $ lein tar
    $ tar xfj target/riemann-0.2.11-SNAPSHOT.tar.bz2
    $ cd riemann-0.2.11-SNAPSHOT

    # with unzip...
    $ unzip -p lib/riemann.jar META-INF/maven/riemann/riemann/pom.properties | sed -n 's/^version=\(.*\)/\1/p'
    0.2.11-SNAPSHOT

    # ...or with bsdtar
    $ bsdtar xfO lib/riemann.jar META-INF/maven/riemann/riemann/pom.properties | sed -n 's/^version=\(.*\)/\1/p' 
    0.2.11-SNAPSHOT

Another option (that's slower to run but doesn't need `unzip` or `bsdtar`) is to parse `pom.properties` from within the `riemann.bin` namespace [like this](https://github.com/aphyr/riemann/compare/master...bfritz:version-display-command).

aphyr:
I think @bfritz's #604 is a cleaner solution to the problem; gonna take his patch for this instead. Thanks all!

PR #553, opened 2015-05-08
Add support for private servers; Migrate to v2 API

jjungnickel:
HipChat can now be installed on private servers. This adds support for this and also switches away from the deprecated v1 API.

### DIFF
--- a/src/riemann/hipchat.clj
+++ b/src/riemann/hipchat.clj
@@ -2,7 +2,6 @@
       :author "Hubert Iwaniuk"}
   riemann.hipchat
   (:require [clj-http.client :as client]
-            [cheshire.core :as json]
             [clojure.string :refer [join]]))
 
 (defn- message-colour [ev]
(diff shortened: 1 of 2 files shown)


src/riemann/hipchat.clj
```diff
@@ -52,14 +51,14 @@
...
                      :room 12345
-                     :from \"Riemann reporting\"
                      :notify 0})]
     (changed-state hc))"
-  [{:keys [token room from notify]}]
+  [{:keys [server token room notify]}]
   (fn [e] (post token
-               {:room_id room
-                :from    from
+               {:server  server
```
aphyr: This will break compatibility for users of the current Hipchat, right? Can we add a sensible (or ...) default for the server?

jjungnickel: Yeah, I'll include that.

jamtur01:
Maybe a line in the docstring explaining which API is supported/deprecated?

jjungnickel:
I don't think that's necessary since private and hosted installations support both v1 and v2 with v1 being deprecated. All that users need to do is get an updated token for v2. I think I'll mention that a v2 key is necessary.

aphyr:
I'll add a note to the changelog about this, but if there's any way you can detect the use of the V1 API and fail at *config* time, that'll probably be helpful for users. :)

jjungnickel:
That's a great idea, I think the keys differ in length/format. I'll investigate that :)

jjungnickel:
Now fails when the key is not the proper size

aphyr:
Looks good to me! Thanks, @jjungnickel. :)

PR #555, opened 2015-05-15
Add InfluxDB 0.9 output

greglook:
Hi @aphyr, we're using Riemann at Counsyl and it's been great so far!

We've been using InfluxDB's new 0.9 series as a metric storage backend. The API changed substantially between 0.8 and 0.9, so the old Capacitor client is not compatible with newer InfluxDB versions. The code below is a tweaked copy of what we're using to batch-write points to influx in production, and it works pretty well. I've also run the (minimal) integration tests against a VM and verified that the data goes through.

Open to feedback on the namespace naming, I just figured it was simplest to suffix it for now. I also considered opening a PR against Capacitor, but this writer is pretty simple and didn't seem like it needs to be a full client.


aphyr:
Cool! I'd really like to avoid a cartesian-product explosion in namespaces x versions though--is there any way we can fold this work into the existing Influx client in a way that does compatibility nicely? That could mean using both the new writer and Capacitor, or maybe making a PR to Capacitor (probably nice for anyone else using Influx, right?). What do you think?

greglook:
>  I'd really like to avoid a cartesian-product explosion in namespaces x versions

Definitely agree! For InfluxDB at least, 0.9 is planned as the next major release and 0.8 is end-of-life, so I'd expect this to replace `riemann.influxdb` at some point. The API incompatibility is annoying since it forces clients like Riemann to explicitly pick a version on either side of the break. :(

One option, we could put these functions into the existing `riemann.influxdb` namespace but suffix the functions instead  (e.g. `influxdb-9`). That way there's one fewer namespace and we just have to rename the functions later. Want me to try that?

greglook:
Also, I think it's fine not to merge this until 0.9 is stable. At the least it'll be a good example for people brave enough to use the InfluxDB RCs. :)

aphyr:
Yeah, let's go ahead and merge these fns into the existing influxdb namespace (cuz presumably this API *will* take effect later). Then at the top level, let's take an option, :version, that specifies whether you'd like to use 0.8 or 0.9. :)

greglook:
Okay @aphyr, take another look. I definitely like this combined approach better.

aphyr:
Oh man, this is phenomenal work Greg. Thank you so much! :)

greglook:
Thanks! Happy to contribute.

PR #557, opened 2015-05-27
Updated slack default-formatter and extended-formatter

pradeepchhetri:
- Added some missing fields like description and tags.


aphyr:
Looks nice. Thanks @pradeepchhetri :)

aphyr:
Just a headsup to also update the relevant tests next time. :)

PR #560, opened 2015-06-01
email: allow sequentials as input

pyr:
This can be useful to avoid useless applies when recipient lists are fetched elsewhere than the raw config.


aphyr:
Thank you, @pyr! :)

pyr:
<3

jamtur01:
Thanks @pyr for making my life easier. :)

PR #561, opened 2015-06-04
Added an example of using a namespaced function inside another namespaced function.

jamtur01:
This is from a somewhat confused conversation on IRC and I may not
totally wrong about how this works. Please feel free not to merge
if a) I am wrong or b) You don't think it's a useful example.

### DIFF
--- a/howto.html
+++ b/howto.html
@@ -2341,6 +2341,29 @@ dig into it here.</p>
   </div>
 </div>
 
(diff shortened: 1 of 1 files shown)


aphyr:
Ah, I think this is already covered by http://riemann.io/howto.html#organizing-with-namespaces

PR #562, opened 2015-06-09
fixing problem with “name” change and rejected null values in influxdb rc31/rc32

mbuczko:
I noticed a problem with recent influxdb 0.9.0 rc31which rejects null values and rc32 where "name" in point structure has been renamed to "measurement". Just in case if find it useful I fixed these two issues.

### DIFF
--- a/src/riemann/influxdb.clj
+++ b/src/riemann/influxdb.clj
@@ -34,6 +34,7 @@
   (let [ignored-fields (set/union special-fields tag-fields)]
     (-> event
         (->> (remove (comp ignored-fields key))
+             (remove #(nil? (val %)))
              (map #(vector (name (key %)) (val %)))
              (into {}))
         (assoc "value" (:metric event)))))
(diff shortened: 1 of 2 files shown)


pradeepchhetri:
@mbuczko You will need to update the influxdb tests also accordingly. Thats why the build is failing.

mbuczko:
@pradeepchhetri influxdb test fixed but build still fails. for some reason it fails even without my patch applied (on clean repo clone). any hint?

pradeepchhetri:
@mbuczko It fails on riemann.transport-test. I tried debugging but even I am not sure why it is failing.

robinroestenburg:
Just my 2cts, but I don't think Riemann should reject null values by default. There are probably a lot more cases of 'wrong' input and it would be impossible to have Riemann take care of them all.

InfluxDB should handle this, there is an issue for this (https://github.com/influxdb/influxdb/issues/3077) and it is planned to be fixed in 0.9.1. 

Thanks for the fixes though @mbuczko, I am happily using your fork together with 0.9.0.

robashton:
I'm not convinced that "it would be impossible to have Riemann take care of them all." covers this.

We use protobuffs to talk to Riemann and while those fields are optional and not sent, they're still materialising as undefined in this code. It'd really be much easier to strip them for the time being :)

aphyr:
Thank you, @mbuczko. :)

Issue #565, opened 2015-06-16
tapped index allows duplicate host/service pairs


mfournier:
```
(let [index (tap :index (index))]
  (streams index))

(tests
  (deftest index-test
    (let [index (:index (inject! [
      {:host "host1" :service "service1" :ttl 10 :time 1}
      {:host "host1" :service "service1" :ttl 10 :time 2}
      {:host "host1" :service "service1" :ttl 10 :time 3}
    ]))]
    (clojure.pprint/pprint index))))
```

```
Testing riemann.config-test
[{:service "service1", :time 1, :host "host1", :ttl 10}
 {:service "service1", :time 2, :host "host1", :ttl 10}
 {:service "service1", :time 3, :host "host1", :ttl 10}]

Ran 1 tests containing 0 assertions.
0 failures, 0 errors.
```

The real index won't allow you to have multiple identical host/service pairs.

In this regard, depending on what/how you test, you'd get a different result than when running riemann for real.

In any case, thanks again for implementing streams testability, it's really super useful to me !

aphyr:
This is by design: taps show you every event that flows through them. They don't change their behavior based on their child streams--and if they did, you wouldn't be able to make assertions about the intermediate state of the index; only the final state.

mfournier:
Thanks for the clarification @aphyr ! It indeed makes sense.

PR #566, opened 2015-06-16
add test/lookup

mfournier:
Writing tests for cases where many events end up in the index can be a
bit cumbersome. This helper function mimics the behaviour of
riemann.index/lookup and hopefully makes writing test a little bit
easier.

### DIFF
--- a/src/riemann/test.clj
+++ b/src/riemann/test.clj
@@ -151,17 +151,17 @@
             persistent!)))))
 
 (defn lookup
-  "Lookup an event in a pseudo-index returned by inject!.
-  The equivalent to index/lookup, for tests.
-  "
-  [index host service]
+  "Lookup an event by host/service in a vector of tapped events returned by
+  inject!. If several matching events have passed through the tap, the last one
+  will be returned."
+  [events host service]
   (last
     (filter
       identity
       (map #(when
               (and (= host (:host %))
                    (= service (:service %))) %)
-           index))))
+           events))))
 
 (defmacro deftest
   "Like clojure.test deftest, but establishes a fresh time context and a fresh
--- a/src/riemann/test.clj
+++ b/src/riemann/test.clj
@@ -150,6 +150,19 @@
                     (transient {}))
             persistent!)))))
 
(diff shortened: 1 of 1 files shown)


mfournier:
Context: when trying to use `riemann.index/lookup` from inside a `(deftest)` block, I get this error: `java.lang.IllegalArgumentException: No implementation of method: :lookup of protocol: #'riemann.index/Index found for class: clojure.lang.PersistentVector`

Maybe this is actually possible, and I simply didn't find how. In this case, we can just drop this patch.

Using this lookup function, I'm able to write tests which look like this:
```clojure
(deftest lookup-demo
  (let [idx (:index (inject! [
    {:host "host1" :service "service1" :state "ok" :ttl 10 :time 1}]))]

    (let [event (lookup idx "host1" "service1")]
      (is (= (:state event) "ok"))
      (is (= (:ttl event) 10)))))
```

As explained in #565, a tap doesn't have exactly the same characteristics as an index. So I'm using `(last)` here to fetch the event in its "final state", to simulate what we would find in the real index. Is this good enough ?

I didn't figure out yet how to test this in `test/riemann/test_test.clj`, but I'll try again if you think this would be a valuable addition.

mfournier:
So I just updated the docstring, as I realised I had a incomplete understanding of how taps could be used. I removed the comparison between a tap and an index which could lead to confusion.

src/riemann/test.clj
```diff
@@ -150,6 +150,19 @@
                     (transient {}))
             persistent!)))))
 
+(defn lookup
+  "Lookup an event by host/service in a vector of tapped events returned by
+  inject!. If several matching events have passed through the tap, the last one
+  will be returned."
+  [events host service]
+  (last
+    (filter
```
aphyr: `(filter identity (map f coll))` can be written as `(keep f coll)`, btw.

aphyr:
I'm not super clear on how often you only want the final state, but it seems useful and the test ns is relatively uncomplicated right now. Sounds good to me! :)

PR #567, opened 2015-06-20
Fixes link to full grammar

viniciushana:


aphyr:
Thank you, @viniciushana! Wouldn't have caught that without your help. :)

PR #569, opened 2015-06-21
Make Influx 0.9 work with latest

robashton:
Influx 0.9 crashes when you send through null fields (No really, completely - it dies a horrible death complete with Go stack traces that put Clojure to shame), there are a ton of paths I can see that end up doing this, so it's easiest to just strip them out at this point in the code.

Also it's no longer "name" in Influx, it's "measurement" so I changed that too.

Same caveat as always, I don't Clojure any more so it's probably horrible.

### DIFF
--- a/src/riemann/influxdb.clj
+++ b/src/riemann/influxdb.clj
@@ -35,6 +35,7 @@
     (-> event
         (->> (remove (comp ignored-fields key))
              (map #(vector (name (key %)) (val %)))
+             (filter (comp #(not (nil? %)) last))
              (into {}))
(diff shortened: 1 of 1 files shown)


robashton:
Caveat more true than usual - I'll set up an actual Clojure dev environment on my laptop and do a proper PR with passing tests instead of just making a half-arsed change to the copy I have lying around on a docker container..

robinroestenburg:
Duplicate of #562 (which fixes the null values and the name=>measurement change).

robashton:
Cool!

PR #572, opened 2015-06-22
fix influxdb 0.9

Jean-Emile:


robinroestenburg:
This work was already done in PR #562.

PR #575, opened 2015-06-26
[WIP] Add first draft of encoding influxdb line protocol

### LINKED ISSUE #574: InfluxDB 0.9 integration broken
Sadly InfluxDB changed their support for JSON during the RC-progress to 0.9. Changed as in just deactivated, so now we need to send line protocol: https://influxdb.com/docs/v0.9/concepts/reading_and_writing_data.html

I will try and come up with a function to encodes events in that very under-defined format.

timbuchwaldt:
This resolves #574 and is currently basically working, altough terribly un-clojurey. I'd love some hints regarding making this more clojure-like.

### DIFF
--- a/src/riemann/influxdb.clj
+++ b/src/riemann/influxdb.clj
@@ -14,6 +14,26 @@
   "A set of event fields in Riemann with special handling logic."
   #{:host :service :time :metric :tags :ttl})
 
+(defn replace-disallowed [field]
+  (clojure.string/replace
+    (clojure.string/replace
+      (clojure.string/replace field " " "\\ ") "=" "\\=") "," "\\,"))
+(defn kv-encode [kv]
+  (clojure.string/join "," (map (fn [[key value]]
+   (str key "=" value)) kv)))
+
+(defn lineprotocol-encode [event]
+  (let [encoded_fields (kv-encode (get event "fields"))
+        encoded_tags  (kv-encode (get event "tags"))]
+
+    (str (get event "name") " " encoded_fields  "\n")))
+
+
+(defn lineprotocol-encode-list [events]
+  ; encode {"points" [{"name" "xyzzy", "time" "2015-06-26T07:06:45.000Z", "tags" {"host" "h"}, "fields" {"value" 0.6514667122989345, "state" "ok", "description" "at 2015-06-26 09:06:45 +0200"}}], "database" "foo"}
+  ; [{"name" "xyzzy", "time" "2015-06-26T07:06:45.000Z", "tags" {"host" "h"}, "fields" {"value" 0.6514667122989345, "state" "ok", "description" "at 2015-06-26 09:06:45 +0200"}}]
+  ; {"name" "xyzzy", "time" "2015-06-26T07:06:45.000Z", "tags" {"host" "h"}, "fields" {"value" 0.6514667122989345, "state" "ok", "description" "at 2015-06-26 09:06:45 +0200"}}
+  (clojure.string/join (pmap lineprotocol-encode (get events "points"))))
 
 (defn event-tags
   "Generates a map of InfluxDB tags from a Riemann event. Any fields in the
@@ -146,7 +166,7 @@
   `:timeout`        HTTP timeout in milliseconds. (default: `5000`)"
   [opts]
   (let [write-url
-        (format "%s://%s:%s/write" (:scheme opts) (:host opts) (:port opts))
+        (format "%s://%s:%s/write?db=%s" (:scheme opts) (:host opts) (:port opts) (:db opts))
 
         payload-base
         (cond->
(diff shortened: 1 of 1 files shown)


aphyr:
Is this already addressed by https://github.com/aphyr/riemann/pull/555?

timbuchwaldt:
Sadly it seems InfluxDB has dropped support for the changes in #555 as they don't support JSON any more. I built on top of those change to re-encode the extracted fields and tags in their new line protocol (which is somewhat undefined)

aphyr:
Cooool, thanks Tim. I'll take a look at this.

robashton:
Oh neat - thanks so much @timbuchwaldt  - This Influx pullava is killing us right now and it's amazing to see that somebody is on the case

timbuchwaldt:
@robashton well.. thats basically as far as my Clojure Skillz (I have none) got me - so I'm happy to see how I could improve on that :)

robashton:
@timbuchwaldt I'd be happy to put my Clojure hat on and have a look at this a little later on today

greglook:
InfluxDB 0.9.0 _does_ support JSON, they just made a backwards-incompatible change to the API. All you need to do is pass `:measurement` instead of `:name` [here](https://github.com/aphyr/riemann/blob/3534463855a08633abf8e7103f75a8ef525a51e7/src/riemann/influxdb.clj#L51).

robinroestenburg:
@greglook true, JSON is still supported in 0.9.x. It is deprecated though and it is scheduled to be removed in one of the upcoming releases.

src/riemann/influxdb.clj
```diff
@@ -14,6 +14,27 @@
...
+        encoded_tags  (kv-encode-9 (get event "tags"))]
+
+    (str (get event "name") "," encoded_tags " " encoded_fields  "\n")))
+
+
+(defn lineprotocol-encode-list-9 [events]
+  ; encode {"points" [{"name" "xyzzy", "time" "2015-06-26T07:06:45.000Z", "tags" {"host" "h"}, "fields" {"value" 0.6514667122989345, "state" "ok", "description" "at 2015-06-26 09:06:45 +0200"}}], "database" "foo"}
+  ; [{"name" "xyzzy", "time" "2015-06-26T07:06:45.000Z", "tags" {"host" "h"}, "fields" {"value" 0.6514667122989345, "state" "ok", "description" "at 2015-06-26 09:06:45 +0200"}}]
+  ; {"name" "xyzzy", "time" "2015-06-26T07:06:45.000Z", "tags" {"host" "h"}, "fields" {"value" 0.6514667122989345, "state" "ok", "description" "at 2015-06-26 09:06:45 +0200"}}
+  (clojure.string/join (pmap lineprotocol-encode-9 (get events "points"))))
```
aphyr: I'd avoid the pmap here; riemann is already using NUM_CORES threads for this work.

aphyr:
Merged #562 instead, which I believe also addresses these problems?

trehn:
@aphyr #562 fixes the problem for now, but not for long. InfluxDB will remove JSON because it's too slow for them. When that happens, you will need to use the line protocol @timbuchwaldt implemented here. So I suggest keeping this PR open, it will have to be merged eventually.

aphyr:
Oh, OK, my bad! I've reopened here--@timbuchwaldt, would you mind dropping the pmap (just wanna keep the executors free cuz we're already running in parallel here), and I'll merge? :)

src/riemann/influxdb.clj
```diff
@@ -14,6 +14,27 @@
   "A set of event fields in Riemann with special handling logic."
   #{:host :service :time :metric :tags :ttl})
 
+(defn replace-disallowed-9 [field]
+  (clojure.string/replace
+    (clojure.string/replace
+      (clojure.string/replace field " " "\\ ") "=" "\\=") "," "\\,"))
```
greglook: A couple suggestions - first, if you require `[clojure.string :as str]` up top you could reduce a lot of the repetition here. Second, this is a perfect example for the thread-first macro:

```clojure
(-> field
    (str/replace " " "\\ ")
    (str/replace "=" "\\=")
    (str/replace "," "\\,"))
```

More generally, the repeated pattern here should be an indication that there's a higher-order function to pull out:

```clojure
(let [special-chars #{" " "=" ","}]
  (reduce #(str/replace %1 %2 (str "\\" %2))
          field special-chars))
```

aphyr: Alternatively: `(str/escape field {\space "\\ ", \= "\\=", \, "\\,"})` :)

greglook: Hah :bow: @aphyr, I forgot about that arity!

aphyr: Wait, there's more than one arity for `escape`?

greglook: Sorry, was thinking `clojure.string/replace`.

timbuchwaldt:
So... thanks for the feedback :) I think I got it all in there, test seems fine too.

src/riemann/influxdb.clj
aphyr: Oh, ya! `replace` works great for patterns but when it's just single chars, `escape` can do the trick! :)

aphyr:
Beautiful, thank you very much, @timbuchwaldt. :)

PR #576, opened 2015-06-30
Fix filter by state

mausch:
I could be wrong as I'm only a newbie to Riemann but it seems that this example is filtering by state, not service.
If I'm mistaken then it might be worth explaining this a bit more as it's quite unintuitive.


aphyr:
Excellent catch--thanks!

PR #578, opened 2015-06-30
Update InfluxDB 0.9.0 writer

### LINKED ISSUE #574: InfluxDB 0.9 integration broken

greglook:
@aphyr fixes #574 

This PR contains two changes. The first is a minor fix for a breaking change introduced by the final 0.9.0 InfluxDB API. For... whatever reason, the `"name"` field changed to `"measurement"`, and `nil` field values are no longer permitted.

The second change is more architectural - rather than specifying a single set of fields to turn into InfluxDB tags, the information is carried in metadata on each event. This way, you can specify a _base set_ of tag fields, but individual events can opt their custom fields into tags. This way consumers can write streams to treat different metrics appropriately, for example declaring that `:response-code` and `:controller` on a website request event should be tags.

### DIFF
--- a/src/riemann/influxdb.clj
+++ b/src/riemann/influxdb.clj
@@ -192,14 +192,12 @@
         tag-fields (:tag-fields opts #{:host})]
     (fn stream
       [events]
-      (let [events (if (sequential? events) events (list events))
-            points (events->points-9 tag-fields events)]
-        (when-not (empty? points)
-          (->> points
-               (assoc payload-base "points")
-               (json/generate-string)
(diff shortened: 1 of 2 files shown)


aphyr:
[I merged #562 instead]

PR #579, opened 2015-07-07
Fix smap example: return a new event instead of a map

nberger:


aphyr:
Just a headsup: Events and maps are interchangeable and are used this way throughout Riemann. Event is just a performance optimization.

nberger:
Thanks! Didn't know that. But in this concrete example, I think `event` is needed so it adds the `:time` field which is needed to construct the body for the `email` ([ref](https://github.com/aphyr/riemann/blob/master/src/riemann/common.clj#L278)). Other downstream processors might need `:time` too. Better to avoid this issue in the how-to. It might be a good idea to explain the difference in other place.

PR #580, opened 2015-07-08
Do not escape slack markup formatting in custom formatter.

zackdever:
This allows custom formatters to make use of slack's markup formatting.
Without this change all text is escaped in custom formatters and the
caller cannot make use of markup formatting (hyperlinking, bold, etc.)


aphyr:
Thanks Zack! :)

PR #582, opened 2015-07-09
Allow graphite to take a function as host name

joerayme:
This allows you to dynamically change the host connected to by each connection in the pool.

For example, our graphite host has multiple IPs in its A record. A function such as this will allow us to spread the connections over several hosts:

```clojure
(defn resolve-host
  "Returns a function which resolves a hostname to one of the IP addresses the record
  returns.

  (resolve-host \"graphite.example.com\")"
  [host]
  (fn []
    (debug "Resolving" host)
    (try (let [env (doto (Properties.) (.put Context/INITIAL_CONTEXT_FACTORY "com.sun.jndi.dns.DnsContextFactory"))
               idc (InitialDirContext. env)
               attrs (.getAttributes idc host (into-array String ["A"]))
               attr (.get attrs "A")]
           (if (not (nil? attr))
             (let [realhost (.get attr (rand-int (.size attr)))]
               (info "Resolved" host "to" realhost)
               realhost)
             host))
         (catch Exception e (error "Exception:" (.getMessage e))))))

(def graph (graphite {:host (resolve-host "graphite.example.com")
                      :port 2003
                      :pool-size 8}))
```

### DIFF
--- a/src/riemann/graphite.clj
+++ b/src/riemann/graphite.clj
@@ -10,7 +10,8 @@
   (:use [clojure.string :only [split join replace]]
         clojure.tools.logging
         riemann.pool
-        riemann.common))
+        riemann.common
+        [riemann.transport :only [resolve-host]]))
 
 (defprotocol GraphiteClient
   (open [client]
@@ -110,12 +111,12 @@
         pool (fixed-pool
                (fn []
                  (info "Connecting to " (select-keys opts [:host :port]))
-                 (let [host (:host opts)
+                 (let [host (resolve-host (:host opts))
                        port (:port opts)
                        client (open (condp = (:protocol opts)
                                       :tcp (GraphiteTCPClient. host port)
                                       :udp (GraphiteUDPClient. host port)))]
-                   (info "Connected")
+                   (info "Connected to" host)
                    client))
                (fn [client]
                  (info "Closing connection to "
--- a/src/riemann/transport.clj
+++ b/src/riemann/transport.clj
@@ -31,7 +31,9 @@
     (io.netty.util.concurrent Future
                               EventExecutorGroup
                               DefaultEventExecutorGroup
-                              ImmediateEventExecutor)))
+                              ImmediateEventExecutor)
+    (java.net InetAddress
+              UnknownHostException)))
 
 (defn ^DefaultChannelGroup channel-group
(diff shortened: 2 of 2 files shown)


aphyr:
Iiiiinteresting. So I feel like this problem should be solved at the resolver layer, and the reason the current strategy doesn't work is because the socket constructor hardcodes addresses[0] as the only IP to use. I think we should instead fix this by adding a function like `riemann.transport/resolve-host` which does the randomized DNS resolution as you've shown here, and changing the graphite client to make that call automatically when it opens a socket. Then everyone gets randomized resolution out of the box. :)

joerayme:
Thanks for your reply. We would be totally happy to add a `riemann.transport/resolve-host` function. However, there are a couple of issues with doing the DNS resolution using the method we described:

- It relies on the hostname being an A record which is fine for us but may not be for everyone (e.g. CNAME)
- it relies on `com.sun.jndi.dns.DnsContextFactory` existing which it may not on all JVMs, and it's required in order to be able to get all addresses for a given A record

If we just use `InetAddress.getAllByName(host)` then the result will be cached and you won't get a different IP address for each pooled connection.

And we don't believe we can even use something like [dnsjava](http://www.dnsjava.org/) to do the lookup because it looks like it just uses `InetAddress.getAllByName(host)` under the hood anyway. In any case that's a pretty heavy library to pull in just to do a tiny resolution.

Any thoughts?

aphyr:
Blerg, this sounds awful. There has *got* to be a better way to handle this than making users write their own DNS resolvers.

aphyr:
InetAddress.getAllByName(host) returns an array of all InetAddresses for the host, right? Even if those values are cached, we can select randomly *from* those addresses to get balanced behavior, ya?

joerayme:
You're absolutely right. I've implemented that and used it in `riemann.graphite`. I haven't added tests as I'm not sure how you'd go about testing something like this. My colleague has previously done something funky with overriding the DNS resolver using a `resources/META-INF/sun.net.spi.nameservice.NameServiceDescriptor` file but I couldn't figure out how to use that in Clojure.

joerayme:
Hey @aphyr, have you had a chance to look at this yet?

aphyr:
Looks great, thank you @joerayme :)

PR #583, opened 2015-07-11
Convert TSDB tags to custom fields

pharaujo:
This also takes advantage of the fact `host` field is a tag, so it sets
it that way. We only need to be sure there isn't a `service` tag so that
it doesn't overwrite the `service` field.


pharaujo:
The build failure seems to be from `riemann.transport-test/tls-test`, not related with the pull request.

aphyr:
Works for me! Thanks pharaujo!

PR #584, opened 2015-07-14
JVM opt for disabling epoll

pharaujo:
There are still cases where a Linux amd64 host can't use the netty epoll
driver - e.g. on CentOS 5, where the glibc version is too old to run it.

This change allows disabling the default and fallback to the pure-java
implementation by adding `-Dnetty.epoll.enabled=false` to JVM opts.


ericfode:
:+1: We need this too at autodesk :) @pharaujo  thank you for making this pull request!

aphyr:
Thanks, @pharaujo!

PR #585, opened 2015-07-21
logging: fix documentation about possible values for layout

vincentbernat:
`json-eventv1` is not a valid layout keyword. Replace with
`json-event-v1`. Also add `json-event-v0`.


pyr:
works for me, sorry for not catching it sooner :+1:

aphyr:
(think this is no longer applicable given the logging rewrite)

PR #586, opened 2015-07-21
logging: fallback to "riemann" when an unknown layout is provided

vincentbernat:
When providing an unknown layout, the user just doesn't get any log.

The usefulness of this PR is quite limited as it doesn't output an error message (too early). The user will be puzzled not getting the log in the expected layout while they were puzzled not getting the logs at all. Not sure what's the best.

### DIFF
--- a/src/riemann/logging.clj
+++ b/src/riemann/logging.clj
@@ -50,7 +50,10 @@
 (defn get-layout
   "Fetch a logging layout by name"
   [layout-name]
-  (get layouts (or layout-name :riemann)))
+  (let [layout (get layouts (or layout-name :riemann))]
+    (when (nil? layout)
+      (binding [*out* *err*] (println "invalid logging layout specified: " layout-name)))
+    (or layout (:riemann layouts))))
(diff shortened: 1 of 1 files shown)


src/riemann/logging.clj
```diff
@@ -50,7 +50,7 @@
 (defn get-layout
   "Fetch a logging layout by name"
   [layout-name]
-  (get layouts (or layout-name :riemann)))
+  (or (get layouts layout-name) (get layouts :riemann)))
```
pyr: This is one of the rare cases where printing to `stderr` could be worthwhile.

```clojure
(let [layout (get layouts layout-name)]
  (when (nil? layout)
      (binding [*out* *err*] (println "invalid logging layout specified: " layout-name)))
  (or layout (:riemann layouts)))
```

vincentbernat:
PR updated with what @pyr proposed (with a slight modification).

aphyr:
Thanks so much vincent! :)

PR #595, opened 2015-08-20
Updating netty to 4.0.30.Final

hiloboy0119:
https://github.com/aphyr/riemann/issues/594


mallman:
Is there a way to rerun these tests? The cause of failure is unclear and the tests pass locally with oracle jdk 1.8.0_60.

aphyr:
This is just Travis being a Travis, I think. Looks good to me too. Thanks! :)

PR #596, opened 2015-08-24
Add tags to opsgenie events

vixns:
Opsgenie use tags for event filtering and bypassing quiet hours. This PR pass event tags to opsgenie events, and use them for computing event alias.

usage exemple: 

      (let [og (opsgenie "your-api-key" "your-recipients")]
        (changed-state {:init "ok"}
          (where (state "ok")
            ; close all tags variants when back to ok
            (with {:tags null} (:resolve og))
            (with {:tags ["Urgent" "OverwriteQuietHours"]} (:resolve og)))
          (where (state "warning") (with {:tags null} (:trigger og)))
          (where (state "critical") 
             (with {:tags ["Urgent" "OverwritesQuietHours"]} (:trigger og)))))


src/riemann/opsgenie.clj
```diff
@@ -40,7 +40,8 @@
   "Generate OpsGenie alias based on event"
   [event]
   (hash (str (:host event) " "
-       (:service event))))
+       (:service event) " "
+       (apply str (:tags event)))))
```
aphyr: Is this hash supposed to be stable for two events with the same tags? Is it supposed to be different when given tags like ["foo" "bar"] vs ["f" "oobar"]? If this is so, can I suggest sorting them and using `clojure.string/join` with a reserved character like `\uffff`?

pyr:
@vixns are you willing to update the PR to reflect the comments above?

vixns:
@pyr updated

pyr:
thank you @vixns

PR #598, opened 2015-08-25
Fix logging, use logback instead log4j

juise:
Related to https://github.com/aphyr/riemann/issues/556

### DIFF
--- a/src/riemann/logging.clj
+++ b/src/riemann/logging.clj
@@ -165,7 +165,7 @@
                 rolling-policy    (doto (TimeBasedRollingPolicy.)
                                     (.setMaxHistory (or rotate-count 10))
                                     (.setFileNamePattern
-                                     (str file ".%d{yyyy-MM-dd}.%i"))
+                                     (str file ".%d{yyyy-MM-dd}"))
                                     (.setParent log-appender)
                                     (.setContext context)
                                     (.start))
--- a/src/riemann/logging.clj
+++ b/src/riemann/logging.clj
@@ -7,6 +7,8 @@
            (ch.qos.logback.core
              ConsoleAppender
              FileAppender)
+           (ch.qos.logback.core.encoder
+             LayoutWrappingEncoder)
            (ch.qos.logback.core.rolling
              RollingFileAppender
              TimeBasedRollingPolicy
@@ -14,6 +16,9 @@
              SizeBasedTriggeringPolicy)
            (ch.qos.logback.classic.encoder
              PatternLayoutEncoder)
+           (net.logstash.logback
+             JSONEventLayoutV0
+             JSONEventLayoutV1)
            (net.logstash.logback.encoder
              LogstashEncoder))
   (:require wall.hack))
@@ -41,36 +46,13 @@
 
 (defmethod encoder :json-event-v0
   [type]
-  (let [hostname (.. java.net.InetAddress getLocalHost getHostName)]
-    (doto (PatternLayoutEncoder.)
-      (.setPattern (str "{\"@fields\":{\"level\":\"%level\","
-                                      "\"threadName\":\"%thread\","
-                                      "\"mdc\":{%mdc},"
-                                      "\"file\":\"%file\","
-                                      "\"class\":\"%class\","
-                                      "\"line_number\":\"%line\","
-                                      "\"method\":\"%method\","
-                                      "\"loggerName\":\"%logger\"},"
-                         "\"@timestamp\":\"%date{yyyy-MM-dd'T'HH:mm:ss.SSS'Z', UTC}\","
-                         "\"@message\":\"%message\","
-                         "\"@source_host\":\"" hostname "\"}\n")))))
+  (doto (LayoutWrappingEncoder.)
+    (.setLayout (JSONEventLayoutV0.))))
 
 (defmethod encoder :json-event-v1
   [type]
- (let [hostname (.. java.net.InetAddress getLocalHost getHostName)]
-   (doto (PatternLayoutEncoder.)
-     (.setPattern (str "{\"level\":\"%level\","
-                        "\"thread_name\":\"%thread\","
-                        "\"mdc\":{%mdc},"
-                        "\"file\":\"%file\","
-                        "\"class\":\"%class\","
-                        "\"line_number\":\"%line\","
-                        "\"method\":\"%method\","
-                        "\"logger_name\":\"%logger\"},"
-                        "\"@timestamp\":\"%date{yyyy-MM-dd'T'HH:mm:ss.SSS'Z', UTC}\","
-                        "\"@version\":1,"
-                        "\"message\":\"%message\","
-                        "\"source_host\":\"" hostname "\"}\n")))))
+  (doto (LayoutWrappingEncoder.)
+    (.setLayout (JSONEventLayoutV1.))))
 
 (defmethod encoder :riemann
   [type]
--- a/src/riemann/logging.clj
+++ b/src/riemann/logging.clj
@@ -1,160 +1,184 @@
 (ns riemann.logging
-  "Configures log4j to log to a file. It's a trap!"
-  ; With thanks to arohner
-  (:import (org.apache.log4j
-             Logger
-             BasicConfigurator
-             EnhancedPatternLayout
+  (:import (org.slf4j
+             LoggerFactory)
+           (ch.qos.logback.classic
              Level
+             Logger)
+           (ch.qos.logback.core
              ConsoleAppender
-             FileAppender
-             SimpleLayout)
-           (net.logstash.log4j JSONEventLayoutV0
-                               JSONEventLayoutV1)
-           (org.apache.log4j.spi RootLogger))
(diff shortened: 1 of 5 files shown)


pyr:
hi @juise, would you willing to look at plugging in https://github.com/pyr/unilog ? we discussed the potential of moving to it with @aphyr but I never got around to it. The syntax is compatible with riemann's

juise:
@pyr looks like the unilog fits very well! If @aphyr doesn't mind I can try to integrate unilog into riemann in next iteration! It may take some time )

@aphyr looks like travis fail somewhere in transport layer, I don't know why, on my desktop all test passed. Could you please take a look?

aphyr:
Yeah, travis-ci fails unreliably on network tests and I can't figure out why. Something changed in their infra. :(

src/riemann/logging.clj
```diff
@@ -1,161 +1,185 @@
...
            (suppress ~more ~@body)
            (finally
              (set-level ~logger old-level#))))
       `(do ~@body))))
 
-(def ^{:doc "available logging patterns"}
-  layouts
-  {:riemann       (EnhancedPatternLayout. "%p [%d] %t - %c - %m%n%throwable")
-   :json-event    (JSONEventLayoutV0.)
-   :json-event-v0 (JSONEventLayoutV0.)
```
aphyr: Do I understand correctly that we're going to break logging for everyone who used json-event, json-event-v0, and json-event-v1? Can we provide compatibiltiy layers?

src/riemann/logging.clj
```diff
@@ -1,161 +1,185 @@
...
+                    otherwise use default time based rotation.
+  :rotate-count     Specifying the number of rotated files to keep. If omitted,
+                    keep last 10 rotated files.
+
+  Layout can be :riemann or :json. If layout omitted, default layout :riemann will be used.
+
   Example:
 
-  (init {:console false :file \"/var/log/riemann.log\"})
+  (init {:console true :file \"/var/log/riemann.log\"})
```
aphyr: Why change this?

src/riemann/logging.clj
```diff
@@ -1,161 +1,185 @@
...
+(defn- get-context
+  []
+  (LoggerFactory/getILoggerFactory))
+
+
+(defmulti encoder identity)
+
+(defmethod encoder :json
+  [type]
+  (LogstashEncoder.))
```
aphyr: Is the Logstash encode the same as json-event-v0 or -v1? Will we break tooling for users who are parsing their JSON logs expecting a particular structure? How does the logstash encoder handle rational coercion?

src/riemann/logging.clj
juise: Unfortunately, you are right, as you can see here (https://github.com/logstash/log4j-jsonevent-layout#v0v1vnothing) the json-event and json-event-v0 is old format, the new format is json-event-v1! The new LogstashEncoder is almost equal with json-event-v1 as you can see below and give more flexibility.

json-event:

{"@fields":{"level":"INFO","threadName":"clojure-agent-send-off-pool-0","mdc":{},"file":"logging.clj","class":"clojure.tools.logging$eval513$fn__519","line_number":"288","method":"invoke","loggerName":"riemann.transport.websockets"},"@timestamp":"2015-08-25T18:30:02.620Z","@message":"Websockets server 127.0.0.1 5556 online","@source_host":"ip-172-31-31-101"}

json-event-v0:

{"@fields":{"level":"INFO","threadName":"clojure-agent-send-off-pool-1","mdc":{},"file":"logging.clj","class":"clojure.tools.logging$eval513$fn__519","line_number":"288","method":"invoke","loggerName":"riemann.transport.websockets"},"@timestamp":"2015-08-25T18:34:22.954Z","@message":"Websockets server 127.0.0.1 5556 online","@source_host":"ip-172-31-31-101"}

json-event-v1:

{"thread_name":"clojure-agent-send-off-pool-1","message":"Websockets server 127.0.0.1 5556 online","@timestamp":"2015-08-25T18:36:29.137Z","level":"INFO","mdc":{},"file":"logging.clj","class":"clojure.tools.logging$eval513$fn__519","line_number":"288","logger_name":"riemann.transport.websockets","method":"invoke","@version":1,"source_host":"ip-172-31-31-101"}

json:

{"@timestamp":"2015-08-25T18:39:24.145+00:00","@version":1,"message":"TCP server 127.0.0.1 2003 online","logger_name":"riemann.transport.tcp","thread_name":"clojure-agent-send-off-pool-4","level":"INFO","level_value":20000,"HOSTNAME":"ip-172-31-31-101"}

Btw, I think it's not truly json, it's only logstash style json. If we want to support truly json, we need to support more functionality from logstash-logback-encoder library.

I suggest to rename :json into :logstash-json or just use the :json keyword as is and break logging backward compatible anyway.

aphyr: Yeah, I'd like it if this PR just changed over to logback without messing around with the data format. I take compatibility pretty seriously and want to give our users a stable base for now.

src/riemann/logging.clj
juise: I find out that the doc strings is a little bit far from reality. So, as you can see I try fix the doc strings and make examples more general. Since other examples below cover different cases with disable console logging, I decide to enable console logging in first example.

aphyr: The reason we chose this is because users most commonly asked how to disable console logging and write purely to a file.

src/riemann/logging.clj
juise: Try to reply on this in the previous comment.

src/riemann/logging.clj
juise: I've got it! But the next example covers this problem, is it?

src/riemann/logging.clj
juise: I spent the last two days in search of a better way to support json-event-v0 / json-events-v1, and found that all of then are terrible. The best of them - compose JSON manually directly from code.

Additionally I suggest to mark this logging scheme as deprecated!

bfritz: @juise Was https://github.com/logstash/logstash-logback-encoder not a good option, at least for json-events-v1?

juise: I tried to use logstash-logback-encoder, but some keys like source_host I can't inject into result output. If you know how can I do it, please let me know!

src/riemann/logging.clj
```diff
@@ -1,161 +1,222 @@
...
+      (.setPattern (str "{\"@fields\":{\"level\":\"%level\","
+                                      "\"threadName\":\"%thread\","
+                                      "\"mdc\":{%mdc},"
+                                      "\"file\":\"%file\","
+                                      "\"class\":\"%class\","
+                                      "\"line_number\":\"%line\","
+                                      "\"method\":\"%method\","
+                                      "\"loggerName\":\"%logger\"},"
+                         "\"@timestamp\":\"%date{yyyy-MM-dd'T'HH:mm:ss.SSS'Z', UTC}\","
+                         "\"@message\":\"%message\","
```
aphyr: What happens when message contains quotes or special chars?

pyr:
It might be worth running a quick poll on the ML to ask who is using `:json-event-v0` there haven't been aggregator releases which leverage that format in a long while (this was specific to older versions of logstash).

An alternative could be to provide the logback code under a new namespace and with a new function name and mark the old logging function as deprecated. This would allow keeping the previous way for a release, making sure nothing breaks and people have a way forward.

juise:
II think we can ask, but I'm not sure we'll get the fast feedback. Btw, I'll try to ask in google group.

I like the alternative way with new (logback) and old (log4j), but it seems complicated to support.

Right now I have no ideas how to do it. I'll try to think about it on weekend.

src/riemann/logging.clj
juise: Fixed! Could you please check?

juise:
I forked log4j-jsonevent-layout and based on it support logback layout! Now we have full support json logging scheme from log4j-jsonevent-layout.

meetmatt:
@aphyr I have tested this pull request on my dev environment and all checks passed
Heavily up-voting for this one! How long will it take to accept this pull request?

aphyr:
Terrific! Sorry bout the delays. Looks like we managed to lose the documentation for which formats exist? I might have missed it--kind of a big patch.

juise:
Could you explain, what kind of documentation we lose? Do you mean documentation about available layouts (:json* and :riemann)?

juise:
One more thing! Should we consider :json-event-v0 and :json-event-v1 as obsolete and mark them in the documentation as deprecated and in the code with (deprecated ....) fun?

P.S. looking forward when you accept this pull request! )

aphyr:
Honestly I can't remember why we added them, haha. I don't use the JSON logging, so I'm happy to take a deprecation patch and if folks complain we can bring em back.

juise:
Awesome! I think I can make the patch with deprecation message in few next days.

Also, I think the issue (https://github.com/aphyr/riemann/issues/551) can be closed as obsolete.

Thanks!

PR #601, opened 2015-09-04
Adds support for UnixNano timestamps in influxdb 0.9

### LINKED ISSUE #600: Sending a UnixNano timestamp causes IntOverflow in InfluxDB
Sending a unix timestamp that counts the nanoseconds since the epoch causes an integer overflow when using `riemann.common/unix-to-iso8601`. I'm using Go, so sending stuff over as a UnixNano timestamp is pretty common and influxdb 0.9 defaults to nanosecond precision, it'd be nice to support unix nano. There was also a small issue where a bunch of our events had the same data in it, but because we were sending a couple thousand events a second and being written to influx, a bunch would have the same timestamp due to the lack of precision, thus overwriting each other.

 Made a patch to fix this, going to make a PR right now.

bg451:
If converting a timestamp from unix to is8601 causes an integer overflow, send the raw timestamp to influxdb. 

Fixes #600

### DIFF
--- a/src/riemann/influxdb.clj
+++ b/src/riemann/influxdb.clj
@@ -127,17 +127,17 @@
 
 
 ;; ## InfluxDB 0.9
-
 (defn event->point-9
   "Converts a Riemann event into an InfluxDB point if it has a time, service,
   and metric."
   [tag-fields event]
   (when (and (:time event) (:service event) (:metric event))
-    {"measurement" (:service event)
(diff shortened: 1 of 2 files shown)


aphyr:
Er, it looks like you may have misunderstood the :time field--it's in POSIX seconds, not nanoseconds. That's why you're getting an integer overflow.

PR #603, opened 2015-09-09
Add ChannelOption/RCVBUF_ALLOCATOR to UDP server

MichaelDoyle:
Without it, packets are truncated at 2048 bytes, regardless of SO_RCVBUF

### DIFF
--- a/src/riemann/transport/udp.clj
+++ b/src/riemann/transport/udp.clj
@@ -97,13 +97,12 @@
                   (.option ChannelOption/SO_BROADCAST false)
                   (.option ChannelOption/MESSAGE_SIZE_ESTIMATOR
                            (DefaultMessageSizeEstimator. max-size))
+                  (.option ChannelOption/RCVBUF_ALLOCATOR
+                           (FixedRecvByteBufAllocator. max-size))
                   (.handler handler))
                 
                 ; Setup Channel options
-                (if (> so-rcvbuf 0)
-                  (doto bootstrap
-                    (.option ChannelOption/SO_RCVBUF so-rcvbuf)
-                    (.option ChannelOption/RCVBUF_ALLOCATOR (FixedRecvByteBufAllocator. so-rcvbuf))))
+                (if (> so-rcvbuf 0) (.option bootstrap ChannelOption/SO_RCVBUF so-rcvbuf))
 
                 ; Start bootstrap
                 (->> (InetSocketAddress. host port)
@@ -163,7 +162,7 @@
          host  (get opts :host "127.0.0.1")
          port  (get opts :port 5555)
          max-size (get opts :max-size 16384)
-         so-rcvbuf (get opts :so-rcvbuf max-size)
+         so-rcvbuf (get opts :so-rcvbuf -1)
          channel-group (get opts :channel-group
                             (channel-group
                               (str "udp-server" host ":" port "(" max-size ")")))
--- a/src/riemann/transport/udp.clj
+++ b/src/riemann/transport/udp.clj
@@ -12,7 +12,8 @@
                              DefaultMessageSizeEstimator
                              ChannelOutboundHandler
                              ChannelInboundHandler
-                             ChannelInboundHandlerAdapter]
+                             ChannelInboundHandlerAdapter
+                             FixedRecvByteBufAllocator]
            [io.netty.channel.group ChannelGroup]
            [io.netty.channel.socket.nio NioDatagramChannel]
            [io.netty.channel.nio NioEventLoopGroup])
@@ -99,7 +100,10 @@
                   (.handler handler))
                 
                 ; Setup Channel options
-                (if (> so-rcvbuf 0) (.option bootstrap ChannelOption/SO_RCVBUF so-rcvbuf))
+                (if (> so-rcvbuf 0)
+                  (doto bootstrap
+                    (.option ChannelOption/SO_RCVBUF so-rcvbuf)
+                    (.option ChannelOption/RCVBUF_ALLOCATOR (FixedRecvByteBufAllocator. so-rcvbuf))))
 
                 ; Start bootstrap
                 (->> (InetSocketAddress. host port)
--- a/src/riemann/transport/udp.clj
+++ b/src/riemann/transport/udp.clj
@@ -163,7 +163,7 @@
          host  (get opts :host "127.0.0.1")
          port  (get opts :port 5555)
          max-size (get opts :max-size 16384)
-         so-rcvbuf (get opts :so-rcvbuf -1)
+         so-rcvbuf (get opts :so-rcvbuf max-size)
          channel-group (get opts :channel-group
                             (channel-group
                               (str "udp-server" host ":" port "(" max-size ")")))


MichaelDoyle:
This should resolve part of the issue reported here:

https://github.com/aphyr/riemann/issues/597

We've seen that UDP packets are truncated after [2048 bytes - which appears to be the default in Netty](https://github.com/netty/netty/blob/4.0/transport/src/main/java/io/netty/channel/socket/DefaultDatagramChannelConfig.java#L46). 

See also: http://stackoverflow.com/a/20392603/2650842

faxm0dem:
could the following be a symptom to this problem?
```
WARN [2015-05-21 12:20:44,953] defaultEventExecutorGroup-2-1 - riemann.transport.udp - UDP handler caught
io.netty.handler.codec.DecoderException: com.google.protobuf.InvalidProtocolBufferException: While parsing a protocol message, the input
 ended unexpectedly in the middle of a field.  This could mean either than the input has been truncated or that an embedded message misr
eported its own length.
        at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:99)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:333)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:319)
        at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:103)
        at riemann.transport.proxy$io.netty.handler.codec.MessageToMessageDecoder$ff19274a.channelRead(Unknown Source)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:333)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:319)
        at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:787)
        at io.netty.channel.nio.AbstractNioMessageChannel$NioMessageUnsafe.read(AbstractNioMessageChannel.java:93)
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:511)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:468)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:382)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:354)
        at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:116)
        at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:137)
        at java.lang.Thread.run(Unknown Source)
```

MichaelDoyle:
Yes, that's exactly what we're seeing.

MichaelDoyle:
While we're looking at this, from an API perspective what is the intention of `:max-size` versus `:so-rcvbuf`. Wondering if these two params are truly independent, or if we want a single unified way to say "this is the max UDP packet size that can be handled".

aphyr:
@MichaelDoyle I'd say we should use max-size to drive so-rcvbuf and whatever other options we need to try to receive messages at that size. Obviously we can't *guarantee* it--the network and OS are allowed to truncate--but the more we can do the better!

Will absolutely merge this, I just need a chance to sit down and do proper PR review, add folks to thank yous, etc. :)

MichaelDoyle:
Ok. Set `so-rcvbuf` to be the same as `max-size` by default. I'm assuming overriding `so-rcvbuf` is useful in its own right, so I left it as an option. But if you think we should drop it entirely, let me know.

faxm0dem:
I don't understand. Setting `so-rcvbuf` to `max-size` means "set the socket's buffer to the size of *one* biggest possible event we want to be able to receive". Sounds insane to me. To me the sanest value for this is the system default, *I.e.* " don't touch it".

MichaelDoyle:
From what I understand, the recieve buffer needs to be _at least_ as large as the biggest possible datagram. If not, any extra bytes will be discarded when the buffer is full. Testing seems to confirm this. Do you have any documentation that suggests otherwise?

faxm0dem:
Here some notes, sorry if you all already know this by heart. Just to make things clear and to explain why I *think* using `max-size == so_rcvbuf` is not a good idea:

Under linux, the *systemctl* parameters that control the behaviour of new UDP sockets regarding `SO_RCVBUF` are `net.core.rmem_default` and `net.core.rmem_max`. The latter controls the maximum buffer size and the latter the default buffer size.

They will control the kernel allocated buffer that will act as a queue behind the application listening on that socket. If the application is slower to handle requests than the incoming rate, the buffer will fill up and the kernel will start to drop the packets.

The usual way of handling this is (see for instance syslog-ng): let the user specify the buffer size, and if (s)he doesn't then don't set it at all and let the kernel set the default value to `net.core.rmem_default`.

On my debian8, the default value is `212992`. On my EL boxes it seems to be `124928`.

Setting the value to `max-size` e.g. to `2048` seems to be incredibly small. As an example on our riemann servers with a few thousand events per second we set it to `4M` which makes the kernel happy. Using the default caused frequent drops (up to 10%).

*Now* from the (tiny) research I did in the netty docs, I can read about `RCVBUF_ALLOCATOR`:

> Allocates a new receive buffer whose capacity is probably large enough to read all inbound data and small enough not to waste its space.

I have no idea what that means. I'm *guessing* this means that netty will adjust change the buffer dynamically according to workload? Well, if it's that, then riemann's `so_rcvbuf` would only be the initial value? Any how, *to be continued*

MichaelDoyle:
Got it. I misunderstood. You think `max-size` will be too small. I agree. From my testing `RCVBUF_ALLOCATOR` does not seem to grow, so I think this needs to be driven off of max-size. That being said, perhaps I've wrongly tied `SO_RCVBUF` and ``RCVBUF_ALLOCATOR` together. I'll take a closer look.

faxm0dem:
It would be great if you could check what netty does, I'm not a java person.

MichaelDoyle:
@faxm0dem I traced this through. Essentially, the `NioDatagramChannel` will use `FixedRecvByteBufAllocator` by default, which will always return a fixed size buffer. This means that by default Netty will not adapt based on the workload (for datagram sockets specifically). 

We have the option of using the `AdaptiveRecvByteBufAllocator`, but I'm not sure why that's not being done already, or what the implications are. For now, it's probably best to just set the buffer size to the max-size.

faxm0dem:
I still don't understand why you want Max=buf. Just leave it alone and les the kernel set it to the default or let user do what he wants

MichaelDoyle:
There are many buffers at play here, so to be precise, this change is with regard to `RecvByteBufAllocator` which is used by the netty library to create a `ByteBuf` to hold data as it's read off the socket. Given this is a concept introduced by netty, there are no kernel level defaults. FWIW, I left the `SO_RCVBUF` socket option alone as the kernel default, or to be specified by the user. Check the diff again, and see if that doesn't make sense?

faxm0dem:
Okay, I'll check. Sorry for the misunderstanding!

faxm0dem:
I'll try out the patch tomorrow to validate that the error messages disappear

aphyr:
So... things here look OK to me! I'm gonna go ahead and merge, but feel free to add follow-on PRs as you see fit. Thanks so much, everyone. :)

PR #606, opened 2015-09-14
fill-in-last*: apply arbitrary function to last event

Anvil:
Hello

I would like to suggest this `fill-in-last*` function. It's basically `fill-in-last`, but it takes a function instead of an update map. The `fill-in-last` function itself can be rewritten using `fill-in-last*`.

Feel free to tell me if I'm taking the wrong path here, but I could not figure out another (rather simple) way to update a field of the last-event depending of *another field* of this last event.

Regards,
Damien.


aphyr:
Seems legit. I'll take it. :)

PR #607, opened 2015-09-18
Xymon: ability to support more message types. Enable/Disable messages implementation

Anvil:
Summary:
* the `format-line` function is renamed as `event->status`, since it converts an event to "status" type messages to Xymon.
* the `xymon` function can take a "`:formatter`" option to send different type of message, but remains compatible with previous implementation.
* the `xymon` and  `send-line` functions can also take a `:timeout` option to make the Xymon socket timeout on both connect and read operations.
* New `event->enable` and `event->disable` functions to create respectively enable and disable type Xymon messages, which allow us to enable and disable Xymon monitoring through Riemann streams.

### DIFF
--- a/src/riemann/xymon.clj
+++ b/src/riemann/xymon.clj
@@ -6,7 +6,7 @@
             [clojure.math.numeric-tower :refer [ceil]])
   (:import java.net.Socket))
 
-(defn format-line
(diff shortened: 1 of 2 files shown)


aphyr:
Cool, thank you!

PR #611, opened 2015-09-27
Fix InfluxDB 0.9 tags

bwilber:
Sorry, I got the escaping wrong for tags, which are different from fields.

### DIFF
--- a/src/riemann/influxdb.clj
+++ b/src/riemann/influxdb.clj
@@ -30,7 +30,7 @@
         encoded_tags   (clojure.string/join "," (map
                          (fn [[tag value]] (str (replace-disallowed-9 tag) "=" (replace-disallowed-9 value)))
                          (get event "tags")))]
-    (str (replace-disallowed-9 (get event "measurement")) "," encoded_tags " " encoded_fields " " (get event "time"))))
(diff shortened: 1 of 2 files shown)


src/riemann/influxdb.clj
```diff
@@ -27,7 +27,9 @@
 
 (defn lineprotocol-encode-9 [event]
   (let [encoded_fields (kv-encode-9 (get event "fields"))
-        encoded_tags   (kv-encode-9 (get event "tags"))]
+        encoded_tags   (clojure.string/join "," (map
+                         (fn [[tag value]] (str (replace-disallowed-9 tag) "=" (replace-disallowed-9 value)))
+                         (get event "tags")))]
     (str (replace-disallowed-9 (get event "measurement")) "," encoded_tags " " encoded_fields " " (get event "time"))))
```
jwilder: Not very familiar with clojure but equals `=` in the measurement name does not need to be escaped.  It does need to be esacped for tag keys an values though.  I _think_ this will escape it inadvertently.

bwilber:
The docs would seem to agree with you.  Thanks!

So, each part of the line protocol has different rules for escaping.

jwilder:
Yes, they are just slightly different unfortunately.

aphyr:
Cool! Thank you both. :)

PR #612, opened 2015-10-06
Fix misplaced docstring in riemann.slack

greglook:
Minor fix here - noticed that some of the Slack functions weren't showing up in the API docs.


aphyr:
Thanks Greg!

PR #614, opened 2015-10-14
VictorOps integration

mallman:
VictorOps integration. We've been using these patches (against 0.2.10) successfully in production for weeks now. Everything has been cool, and we'd like to contribute these back to the Riemann project.

### DIFF
--- a/src/riemann/config.clj
+++ b/src/riemann/config.clj
@@ -41,6 +41,7 @@
                      [test        :as test :refer [tap io tests]]
                      [time        :refer [unix-time linear-time once! every!]]
                      [twilio      :refer [twilio]]
+                     [victorops   :refer [victorops]]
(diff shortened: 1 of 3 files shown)


aphyr:
Cool! Could you do me a favor and add (or replace the integration tests) with tests that mock the call to clj-http? I'm trying to get all the integration libraries set up so they can be run in CI and so forth without embedding keys in the repo. :)

See https://github.com/aphyr/riemann/blob/master/test/riemann/keenio_test.clj#L9 for an example!

mallman:
Sure. I'm no clojure expert, but I'll give it a go. Cheers.

mallman:
Please retest. It looks like one of the builds timed out.

aphyr:
Looks good! Thanks @mallman! :)

PR #615, opened 2015-10-16
Xymon: fixes, scalability, multiple xymon host, error handling

Anvil:
Hello !

Here is a new bunch of updates for the riemann/xymon module which brings :
* Ability to send combo messages (multiple messages in a single TCP connection), through the new ```xymon-batch``` function. This should improve scalability and also makes more sense in statements such as ```(batch (xymon ...))```)
* Ability to send messages to multiple xymon hosts without too much pain.
* Ability to define per-xymon-server connection error handlers.
* A few minor bug fixes

Feel free to comment on the code quality, if need be.

Regards,
Damien.

### DIFF
--- a/src/riemann/xymon.clj
+++ b/src/riemann/xymon.clj
@@ -70,31 +70,31 @@
     (format "disable %s.%s %s %s"
             host service (int (ceil (/ ttl 60))) description)))
 
-(defn *send-line-error-handler*
+(defn *send-message-error-handler*
   "
   Logs given exception as an error message.
 
-  *send-line-error-handler* is the default error handler invoked by
-  send-line if none is provided.
+  *send-message-error-handler* is the default error handler invoked by
+  send-single-message if none is provided.
   "
   [opts exception]
   (error exception (format "cannot reach xymon host (%s:%s)"
                            (:host opts) (:port opts))))
 
-(defn send-line
+(defn send-single-message
   "
-  Connects to Xymon server, sends line, then closes the connection.
-  This is a blocking operation and should happen on a dedicated
-  thread.
+  Connects to a Xymon server, sends message, then closes the
+  connection. This is a blocking operation and should happen on a
+  dedicated thread.
 
   If any exception is raised during the connect/send process, the
-  result of (:error-handler opts *send-line-error-handler*) is invoked
-  as a function with the exception as its parameter.
+  result of (:error-handler opts *send-message-error-handler*) is
+  invoked as a function with opts and the exception as its parameter.
   "
-  [opts line]
+  [opts message]
   (let [opts (merge
               {:host "127.0.0.1" :port 1984 :timeout 5
-               :error-handler *send-line-error-handler*}
+               :error-handler *send-message-error-handler*}
               opts)]
     (try
       (let [addr (InetSocketAddress. (:host opts) (:port opts))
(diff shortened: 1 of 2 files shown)


Anvil:
(I will update test suite next week, btw)

src/riemann/xymon.clj
```diff
@@ -70,40 +70,117 @@
...
-          sock (Socket.)]
-      (.setSoTimeout sock (:timeout opts))
-      (.connect sock addr (:timeout opts))
-      (with-open [writer (io/writer sock)]
-        (.write writer line)
-        (.flush writer)))
-    (catch Exception e
-      ((:error-handler opts) e))))
+(defn *send-message-error-handler*
+  "
```
aphyr: Just as a style thing, we usually start the docstring on the same line as the quote. :)

src/riemann/xymon.clj
```diff
@@ -70,40 +70,117 @@
...
-          addr (InetSocketAddress. (:host opts) (:port opts))
-          sock (Socket.)]
-      (.setSoTimeout sock (:timeout opts))
-      (.connect sock addr (:timeout opts))
-      (with-open [writer (io/writer sock)]
-        (.write writer line)
-        (.flush writer)))
-    (catch Exception e
-      ((:error-handler opts) e))))
+(defn *send-message-error-handler*
```
aphyr: *foo* usually denotes a dynamic variable, but this looks like a function--maybe just call it send-message-error-handler? Or maybe just send-message-error?

src/riemann/xymon.clj
```diff
@@ -70,40 +70,117 @@
...
+  Logs given exception as an error message.
+
+  *send-message-error-handler* is the default error handler invoked by
+  send-single-message if none is provided.
+  "
+  [opts exception]
+  (error exception (format "cannot reach xymon host (%s:%s)"
+                           (:host opts) (:port opts))))
+
+(defn send-single-message
```
aphyr: If you're doing TCP socket manipulation, can I suggest using riemann.pool similarly to how riemann.graphite works? It'd give you thread-safe re-use of sockets and automatic connection management.

src/riemann/xymon.clj
```diff
@@ -70,40 +70,117 @@
...
+  "
+  Turns events into combo messages (using events->combo) and send them
+  to Xymon.
+  "
+  [opts events]
+  (let [formatter (or (:formatter opts) event->status)]
+    (doseq [combo-message (events->combo formatter events)]
+      (send-message opts combo-message))))
+
+(defn xymon-batch
```
aphyr: It might be convenient for the xymon adapter to just accept collections of events and automatically batch them using this function. Maybe check the argument using (sequential? x) to see if it's a list or vector?

src/riemann/xymon.clj
Anvil: Noted. Fixed !

src/riemann/xymon.clj
Anvil: You're right. It's fixed.

src/riemann/xymon.clj
Anvil: I'll dig into riemann.pool, thanks for the hint.

src/riemann/xymon.clj
Anvil: Indeed. I've merged xymon and xymon-batch and do some collateral cleaning.

Anvil:
Hello Kyle ! (HNY btw :] )
It's been a while and I don't think I will have time to integrate riemann.pool in the riemann.xymon client  before Spring. The code in this pull request has been integrated, installed in our production environment and seems to be stable so far.
May I suggest that we merge this pull request, and that as soon as I have some time to do it, I will rework the socket part and submit another pull request ?
Thanks in advance.

PR #619, opened 2015-10-29
Bump the nREPL dep to 0.2.11

bbatsov:
Some tools (e.g. CIDER) rely on the source-tracking evaluation support introduced in it.


aphyr:
Thanks!

PR #625, opened 2015-11-21
Allow overriding graphite metric conversion method

agile:
I noticed some external verifications of what riemann was sending to graphite were failing due to lost precision of some `long` metrics because they're being converted into floats. I'm not sure if there's a good reason to use float vs double so I thought to allow overriding the function used to do the conversion.

### DIFF
--- a/src/riemann/graphite.clj
+++ b/src/riemann/graphite.clj
@@ -79,7 +79,8 @@
 (defn graphite-metric
   "convert riemann metric value to graphite"
   [event]
-  (float (:metric event)))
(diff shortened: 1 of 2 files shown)


aphyr:
IIRC the problem here is fractions--Riemann has a rational type, and Graphite doesn't. How about doing an (integer?) check and passing those through unchanged, and coercing other types to (double x)?

agile:
@aphyr sounds sensible to me, updated to do that.

aphyr:
Looks great, thank you!

PR #626, opened 2015-11-24
Fix percentiles documentation for correct service name

jamtur01:
The current percentiles documentation suggest that the service name is changed like so:

    resource name

becomes

    resource name .95

In reality it becomes:

    resource name 0.95

This change updates the docs to reflect reality.


aphyr:
Thanks!

PR #627, opened 2015-11-27
Add support for KairosDB HTTP integration and metric TTLs

thorstadt:
Currently, the Riemann KairosDB integration only supports the KairosDB telnet/TCP interface, which doesn't offer support for the per-data-point TTL feature. This PR adds the ability to use the KairosDB HTTP API, which does allow you to specify TTLs for every data point. TCP will remain the default.

Also, since the KairosDB HTTP endpoint is considerably slower than the TCP endpoint, this also adds optional support for batching metrics to KairosDB to help overcome the significant latency that results from making individual add data point calls over HTTP.

### DIFF
--- a/src/riemann/kairosdb.clj
+++ b/src/riemann/kairosdb.clj
@@ -8,8 +8,7 @@
              InetAddress)
    (java.io Writer OutputStreamWriter))
   (:require [clj-http.client :as client]
-            [cheshire.core :as json]
-            [riemann.streams :refer [batch where]])
+            [cheshire.core :as json])
   (:use [clojure.string :only [split join replace]]
         clojure.tools.logging
         riemann.pool
@@ -92,9 +91,9 @@
     :http (open (KairosDBHTTPClient. host port))))
 
 (defn kairosdb
-  "Returns a function which accepts an event and sends it to KairosDB.
-  Silently drops events when KairosDB is down. Attempts to reconnect
-  automatically every five seconds. Use:
+  "Returns a function which accepts a single event or collection of events
+  and sends them to KairosDB. Silently drops events when KairosDB is down.
+  Attempts to reconnect automatically every five seconds. Use:
 
   (kairosdb {:host \"kairosdb.local\" :port 4242 :protocol :tcp})
 
@@ -128,14 +127,7 @@
   :ttl                  A function which, given an event, returns the TTL in
                         seconds.
                         Note: TTL is only supported in the HTTP API, and is
-                              ignored when sent via Telnet.
-
-  :batch                Batch metrics into KairosDB. Consider using this when
-                        :protocol is :http. Default false.
-
-  :batch-opts           Map containing :n and :dt, which correspond to Riemann's
-                        `streams/batch` parameters. Used only when :batch is true.
-                        Default {:n 100 :dt 1}"
+                              ignored when sent via Telnet."
   [opts]
   (let [opts (merge {:host "127.0.0.1"
                      :port 4242
@@ -144,9 +136,7 @@
                      :pool-size 4
                      :tags kairosdb-tags
                      :metric-name kairosdb-metric-name
-                     :ttl (constantly 0)
-                     :batch false
-                     :batch-opts {:n 100 :dt 1}} opts)
+                     :ttl (constantly 0)} opts)
         pool (fixed-pool
               (fn []
                 (info "Connecting to " (select-keys opts [:protocol :host :port]))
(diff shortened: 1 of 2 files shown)


src/riemann/kairosdb.clj
```diff
@@ -75,7 +123,19 @@
...
   :block-start          Wait for the pool's initial connections to open
-                        before returning."
+                        before returning.
+
+  :ttl                  A function which, given an event, returns the TTL in
+                        seconds.
+                        Note: TTL is only supported in the HTTP API, and is
+                              ignored when sent via Telnet.
+
+  :batch                Batch metrics into KairosDB. Consider using this when
```
aphyr: Instead of making these arguments to every client, just have your stream accept both single events and vectors of events. Then users can re-use `(batch)` everywhere. :)

thorstadt: Ah, hat's a good idea -- seems obvious in retrospect :-). See abcbf85bdea86dcccc402912a1b0afbadfce2b92

src/riemann/kairosdb.clj
```diff
@@ -75,7 +122,12 @@
                         the pool. Default 0.1.
 
   :block-start          Wait for the pool's initial connections to open
-                        before returning."
+                        before returning.
+
+  :ttl                  A function which, given an event, returns the TTL in
```
aphyr: :ttl is already a field on each event, so do we need to override it here? Would the existing stream transforms like (with :ttl 5 ...) do the trick?

thorstadt: So if I understand, you're proposing using the Riemann event TTL as the KairosDB metric TTL.

I think it's likely that most people using the KairosDB integration won't want TTLs on their metrics since this is a new and less commonly used feature in KairosDB, and it could be very surprising for people to have their metrics start disappearing from KairosDB due to Riemann TTLs leaking into them.

However, I don't particularly like the idea of passing a function to assign TTLs when we could let users do something perhaps more idiomatic like `(with :kairos/TTL 5 ...)`. That way we don't run the risk of leaking Riemann TTLs, and we can do away with the awkwardness of this TTL function business that admittedly feels like it's going against the grain of Riemann. Does that make sense?

aphyr: On 01/09/2016 09:32 PM, Ryan Crum wrote:
> In src/riemann/kairosdb.clj
> <https://github.com/aphyr/riemann/pull/627#discussion_r49271852>:
>
>> @@ -75,7 +122,12 @@
>>                          the pool. Default 0.1.
>>
>>    :block-start          Wait for the pool's initial connections to open
>> -                        before returning."
>> +                        before returning.
>> +
>> +  :ttl                  A function which, given an event, returns the TTL in
>
> So if I understand, you're proposing using the Riemann event TTL as the KairosDB
> metric TTL.
>
> I think it's likely that most people using the KairosDB integration won't want
> TTLs on their metrics since this is a new and less commonly used feature in
> KairosDB, and it could be very surprising for people to have their metrics start
> disappearing from KairosDB due to Riemann TTLs leaking into them.
>
> However, I don't particularly like the idea of passing a function to assign TTLs
> when we could let users do something perhaps more idiomatic like |(with
> :kairos/TTL 5 ...)|. That way we don't run the risk of leaking Riemann TTLs, and
> we can do away with the awkwardness of this TTL function business that
> admittedly feels like it's going against the grain of Riemann. Does that make sense?

Ahhhh so TTL here is when we start destroying data, not the duration of validity 
for the event? Yeah, that's semantically different from Riemann's TTL, and I'd 
say having a function like you've done here is the right call. :)

--Kyle

thorstadt: Cool :+1:

dbriones:
Anything holding up this PR? Would love to see this in the next release...

src/riemann/kairosdb.clj
```diff
@@ -15,26 +17,59 @@
 (defprotocol KairosDBClient
   (open [client]
         "Creates a KairosDB client")
-  (send-line [client line]
-        "Sends a formatted line to KairosDB")
+  (send-metrics [client metrics]
+                "Sends a collection of metrics to KairosDB")
```
jamtur01: Extra tab?

thorstadt: @jamtur01 Not really an extra tab -- just `clojure-mode` indentation, which indents the docstring with the start the args vector.

dbriones:
Bump

jamtur01:
Thanks @thorstadt

PR #628, opened 2015-12-01
added querying data capabilities to influxdb

ivanB1975:
I added querying capabilities to the influxdb driver. I tested both the 0.8 than the 0.9 version of the database. Tests execute correctly. I hope this is a useful modification.
Bests
Ivan

### DIFF
--- a/src/riemann/influxdb.clj
+++ b/src/riemann/influxdb.clj
@@ -1,5 +1,5 @@
 (ns riemann.influxdb
-  "Forwards events to InfluxDB. Supports both 0.8 and 0.9 APIs."
+  "Forwards and get events to/from InfluxDB. Supports both 0.8 and 0.9 APIs."
   (:require
     [capacitor.core :as capacitor]
     [cheshire.core :as json]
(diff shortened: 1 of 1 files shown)


aphyr:
I'm... not sure what this is for, exactly. This adapter exists to turn Riemann events into InfluxDB data points. Would users submit events to perform queries? Why would they want the query results thrown away? Are there side effects involved?

ivanB1975:
I need to store in influxdb events data from Riemann and I need to be able to read from time to time from influxdb to execute a calculation. I need to go back in time to execute the calculation. The results are also stored in influxdb.
Adding the possibility to read from influxdb helped me to achieve this goal.
Generalizing you can think to an use case where a programmer needs to monitor a service for a month. Events containing metric from the service are written to influxdb on a regular basis. The real time monitoring is not enough and the programmer needs to execute a calculation every day on points written also 1 month earlier. 
So it is why you would have to query about results in the past. I think that this use case could justify the additional capability to read from influxdb. The capability to process older data to calculate trends or probabilities for example.

aphyr:
Okay, but we already have an influxdb client, right? Couldn't you just call its querying capabilities directly? I don't think it's appropriate for querying to be a *stream*.

ivanB1975:
Ah but I am doing this, am I? I use an if at line 180. If the method is GET it doesn't return a stream but a map parsed from the body. So I use the client in two ways. If the method is a POST it returns a stream as it was before, while if it is a GET it returns a map.

aphyr:
Having one function do two completely unrelated things isn't something I'm willing to accept. Just call the client's GET request directly.

PR #630, opened 2015-12-02
Extend influxdb client to support changing of precision and measurements

royalaid:
Influxdb client can now take in :precision and :measurement  for the
0.9 client when those parameters need to be changed. It will act the
same as before when they are not supplied.


src/riemann/influxdb.clj
```diff
@@ -151,13 +158,11 @@
   `:tag-fields`     A set of event fields to map into InfluxDB series tags.
                     (default: `#{:host}`)
   `:tags`           A common map of tags to apply to all points. (optional)
-  `:timeout`        HTTP timeout in milliseconds. (default: `5000`)"
+  `:timeout`        HTTP timeout in milliseconds. (default: `5000`)
+  `:measurement`    The name of the measurement to be used by influx
```
aphyr: The docs indicate :measurement is an argument to the client but it looks like instead you're using :measurement from the *event*? I'm a little confused about what's going on here.

royalaid: The :measurement is similar to series from influx 0.8 and it should be on a per event basis. Should I rework the docs?

src/riemann/influxdb.clj
```diff
@@ -118,12 +118,19 @@
...
+            (str "&rp=" (:retention opts))))) 
+
 (defn event->point-9
   "Converts a Riemann event into an InfluxDB point if it has a time, service,
   and metric."
   [tag-fields event]
-  (when (and (:time event) (:service event) (:metric event))
-    {"measurement" (:service event)
+  (when (and (:time event) (or (:measurement event) (:service event)) (:metric event))
+    {"measurement" (:measurement event (:service event))
```
pyr: To make it more evident what you are doing here, please use:

```clojure
(or (:measurement event) (:service event))
```

src/riemann/influxdb.clj
```diff
@@ -151,13 +158,11 @@
   `:tag-fields`     A set of event fields to map into InfluxDB series tags.
                     (default: `#{:host}`)
   `:tags`           A common map of tags to apply to all points. (optional)
-  `:timeout`        HTTP timeout in milliseconds. (default: `5000`)"
+  `:timeout`        HTTP timeout in milliseconds. (default: `5000`)
+  `:measurement`    The name of the measurement to be used by influx
+  `:precision`      The precision (n,u,ms,s,m,h) of the timestamp to be sent to influxdb (default: `s`)"
```
pyr: Why allow precision to be specified if there are no facilities to convert seconds to the other precisions?

jamtur01:
@royalaid Any update on the feedback? Thanks!

src/riemann/influxdb.clj
royalaid: Merely exposes the influxdb option so if you want to scale it for some reason you can.

jamtur01:
Closing as stale. Please re-open if you get a chance! Thanks!

PR #632, opened 2015-12-10
Post to HipChat in JSON format, and ensure 'notify' field is a boolean.

ophelan:
HipChat Server API may have changed to require explicit boolean; this was causing an API error and undelivered events. This changes the post to comply with the API specs, so everybody is happy. Tests pass on both hosted and server versions.

### DIFF
--- a/src/riemann/hipchat.clj
+++ b/src/riemann/hipchat.clj
@@ -42,6 +42,7 @@
   [token {:keys [server room_id] :as conf} event]
   (client/post (str (chat-url server room_id) "?auth_token=" token)
                {:form-params           (format-event (assoc conf :message_format "text") event)
+                :content-type          :json
                 :socket-timeout        5000
                 :conn-timeout          5000
                 :accept                :json
@@ -70,5 +71,5 @@
   (fn [e] (post token
                 {:server  (or server "api.hipchat.com")
                  :room_id room
-                 :notify  notify}
+                 :notify  (boolean notify)}
                e)))


aphyr:
Soooo what was being passed before? null? If folks were passing strings like "true" or "false" do we need to map those as well?

ophelan:
Without the change, the request is application/x-www-form-urlencoded and appears as: "message=Host%3A+appdev01-nops-us-east-1a+%0AService%3A+servers.loadavg.01+%0AState%3A+warning+%0AMetric%3A+3.22+%0ADescription%3A+Riemann+testing+-+disregard&notify=0&room_id=2248597&server=api.hipchat.com&message_format=text&color=yellow" and receives an API error in response: "Value u'false' for field 'notify' is not of type 'boolean'\"

The error occurs regardless of the value of :notify, as well as with :notify unspecified.

Changing to application/json fixes the issue: "{\"message\":\"Host: appdev01-nops-us-east-1a \\nService: servers.loadavg.01 \\nState: warning \\nMetric: 2.13 \\nDescription: Riemann testing - disregard\",\"notify\":true,\"room_id\":2248597,\"server\":\"api.hipchat.com\",\"message_format\":\"text\",\"color\":\"yellow\"}"

That said, it appears that the boolean function returns true for both 0 and 1 (https://clojuredocs.org/clojure.core/boolean), so that's no good. Setting :notify to "true" or "false" works as intended.

aphyr:
Yeah, in Clojure everything but `false` and `nil` are considered logical true. I... I don't think we should break that principle here. It looks like the problem is actually *documentation*: the docstring tells people to pass `0` instead of `false`. What about updating the documentation and adding an (assert (boolean? notify))?

ophelan:
I agree, sticking with the standard nomenclature makes sense.

It gets more interesting. There appears to be a difference in the HipChat SaaS platform and the self-hosted version. The SaaS platform accepts basically any value for the :notify field without error. The self-hosted version returns an error for :notify = 0, false, and "false" when submitting via the urlencoded method. So, if you think changing the post to json is a bad move, I'll take this up with HipChat since it seems like there's a bug there.

"Value u'0' for field 'notify' is not of type 'boolean'\"
"Value u'false' for field 'notify' is not of type 'boolean'\"
"Value u'false' for field 'notify' is not of type 'boolean'\"

aphyr:
Huh, if notify `false` doesn't work, how *are* you supposed to tell it not to notify? I agree, this kinda sounds like an upstream bug?

aphyr:
Gonna close this for now, but if you follow up with upstream and still need to make changes to the client let me know!

PR #633, opened 2015-12-12
Updated RPM to use systemd to remove daemonize dependency

jamtur01:
- Added pkg/rpm/riemann.service file
- Supports start, stop, reload

This deprecates the previous init.sh script that required daemonize.

This assumes your Red Hat-family system supports systemd.


jamtur01:
Ping @mfournier for comment.

faxm0dem:
as many folks still use EL6, the leiningen part would probably need a switch or autodetect

aphyr:
If we need to support both EL6 and EL7 then I suggest we do that switching *inside* the package instead of as a lein task--I'd rather not add another build target. Could we use a postinstall script to detect systemd and copy the right script?

pkg/rpm/riemann.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Riemann
+After=network.target
+
+[Service]
+User=riemann
+EnvironmentFile=/etc/sysconfig/riemann
+ExecStart=/usr/bin/riemann /etc/riemann/riemann.config
+ExecReload=/bin/kill -HUP ${MAINPID}
+ExecStop=/bin/kill -WINCH ${MAINPID}
```
wickedshimmy: `WINCH`?

mfournier:
IMHO the problem here is not systemd, but the dependency on daemonize.

We can bundle both the sysv initscript and the systemd unit file in the same RPM. And ideally have a conditional postinst script which runs `chkconfig` or `systemctl daemon-reload` depending on the distro version.

But if we want to drop the dependency on daemonize, I think we should change the initscript for ante-systemd distros, and make it fail with a useful error message ("please install daemonize from EPEL6" or whatever).

pkg/rpm/riemann.service
jamtur01: Old habits.

jamtur01:
@mfournier Or rewrite the init script to just not use daemonize?

pyr:
So @mfournier @jamtur01 do we keep this around or close this in anticipation of a follow-up PR dropping the daemonize requirement from init scripts?

mfournier:
daemonize is almost in epel7 ! Please post a couple of upvotes here: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2015-74071f9932 ;)

So this is on tracks for real @pyr, this issue can be closed as far as I'm concerned.

pyr:
@mfournier thanks!

PR #637, opened 2016-01-02
Added IRC plugin

jamtur01:
Sends messages from Riemann to an IRC server.

Configured via:

    (def sendirc (irc "chat.freenode.net", 6667, "riemann-bot", "#riemannbot"))

1. Connects to IRC server and stays connected (because IRC connection
takes forever).
2. Waits for messages.
3. Sends formatted events as PRIVMSGs to the specified channel.

This is super crude using Socket and my Clojure probably leaves a lot to be desired.

### DIFF
--- a/src/riemann/irc.clj
+++ b/src/riemann/irc.clj
@@ -1,64 +1,106 @@
 (ns riemann.irc
-  "Forwards events to IRC"
-  (:use [clojure.string :only [join upper-case]])
-  (:import (java.net Socket)
-           (java.io PrintWriter InputStreamReader BufferedReader)))
+  "Forwards events to IRC."
+  (:refer-clojure :exclude [replace])
+  (:import
+   (java.net Socket
+             DatagramSocket
+             DatagramPacket
+             InetAddress)
+   (java.io Writer OutputStreamWriter))
+   (:use [clojure.string :only [join upper-case]]
+        clojure.tools.logging
+        riemann.pool
+        riemann.common
+        [riemann.transport :only [resolve-host]]))
+
+(defprotocol IrcClient
+  (open [client]
+        "Creates an IRC client")
+  (login [client nick channel]
+         "Logs into IRC")
+  (send-line [client channel line]
+        "Sends a formatted line to IRC")
+  (close [client]
+         "Cleans up (closes sockets etc.)"))
+
+(defrecord IrcConnection [^String host ^int port]
+  IrcClient
+  (open [this]
+    (let [sock (Socket. host port)]
+      (assoc this
+             :socket sock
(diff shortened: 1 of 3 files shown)


src/riemann/irc.clj
```diff
@@ -0,0 +1,64 @@
...
+  "Creates an adapter to forward events to IRC. The IRC event will
+  contain the host, state, service, metric and description.
+
+  Requires: server, port, name of IRC user to send events, and channel name.
+
+  Can be configured like so:
+
+  (def sendirc (irc \"chat.freenode.net\", 6667, \"riemann-bot\", \"#riemannbot\"))"
+  [server port nick channel]
+  (def irc-connection (connect server port))
```
aphyr: You probably don't want to define irc-connection as a global here, nor do I think you only want to connect once--it should automatically reconnect. Take a look at how graphite-client uses riemann.pool for automatic connection management.

jamtur01:
@aphyr I think that's much improved.

jamtur01:
Hold off on merging. I need to test re-connection and connection some more.

src/riemann/irc.clj
```diff
@@ -0,0 +1,104 @@
...
+             DatagramPacket
+             InetAddress)
+   (java.io Writer OutputStreamWriter))
+   (:use [clojure.string :only [join upper-case]]
+        clojure.tools.logging
+        riemann.pool
+        riemann.common
+        [riemann.transport :only [resolve-host]]))
+
+(defprotocol IrcClient
```
aphyr: Since you only have one implementation, you don't really have to go to the trouble of defining a protocol and defrecord. A few functions that take a map like {:socket ... :writer ...} could do the trick, I think.

pyr:
@jamtur01 if you want you can re-open the PR on a branch you'd create against this repo, that way I'd be able to lend a hand on the PR

AeroNotix:
Is there not an IRC client in Clojure?

pyr:
@AeroNotix there are a few.

AeroNotix:
Wouldn't it be better to look into adding a Clojure IRC library rather than adding (limited) bespoke support for IRC

pyr:
I would be very favorable to a netty4 based client, to limit the additional burden on the classpath for riemann. That being said, that's not the route @jamtur01 has taken with his approach and he's the one who put in the work to submit a PR, unless we're willing to do the same, our arguments won't hold :-)

jamtur01:
I intend to refactor this at some point but time is not my friend right now.

jamtur01:
Going to close this for now. I don't have time to fix it. Maybe in the summer.

PR #639, opened 2016-01-06
Added support for a PagerDuty formatter for events

jamtur01:
This adds support for passing in a function name for a formatter for PagerDuty events. I suspect there's almost certainly a better way of doing this. Happy to take input on a better implementation. This was cut quickly to provide me with a solution.


aphyr:
Cool beans, thanks James!

PR #640, opened 2016-01-08
close channel when it becomes unwritable

### LINKED ISSUE #623: riemann stops working properly following OOME
Hello,
I am running riemann 0.2.9 inside a docker container with java version 

```
java version "1.7.0_75"
OpenJDK Runtime Environment (IcedTea 2.5.4) (7u75-2.5.4-1~trusty1)
OpenJDK 64-Bit Server VM (build 24.75-b04, mixed mode)
```

Every week or so, my riemann server stops working properly following OOME exception. I got the following stderr output:

```
Exception in thread "riemann task 3" Exception in thread "riemann task 0" Exception in thread "Thread-7" 
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "riemann task 0"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "riemann task 3"
java.lang.OutOfMemoryError: Java heap space
Exception in thread "Thread-6" java.lang.OutOfMemoryError: Java heap space
Exception in thread "epollEventLoopGroup-4-1" java.lang.OutOfMemoryError: Java heap space
Exception in thread "epollEventLoopGroup-4-3" java.lang.OutOfMemoryError: Java heap space
Exception in thread "defaultEventExecutorGroup-2-2" java.lang.OutOfMemoryError: Java heap space
Exception in thread "defaultEventExecutorGroup-2-1" java.lang.OutOfMemoryError: Java heap space
```

and lot of the following lines in riemann's own logs:

```
WARN [2015-11-15 14:28:02,021] epollEventLoopGroup-4-2 - io.netty.channel.DefaultChannelPipeline - Failed to submit an exceptionCaught() event.
WARN [2015-11-15 14:28:02,022] epollEventLoopGroup-4-2 - io.netty.channel.DefaultChannelPipeline - The exceptionCaught() event that was failed to submit was:
WARN
```

I would expect riemann to simply stop working in case of OOME, but I may be misunderstanding something. What should I do to prevent this from happening? Host has 4GB of RAM and data input is really not important (collectd stats every minute or so plus logs from our app which amounts to maximum of 5 messages/sec). 

Thanks for any advice.

abailly:
this happens when channel's  buffer is filled with pending messages because client
does not read them fast enough (or at all). This should prevent OOME from occuring
because of misbehaving clients. This should fix #623

### DIFF
--- a/src/riemann/transport/tcp.clj
+++ b/src/riemann/transport/tcp.clj
@@ -54,6 +54,11 @@
   ChannelInboundHandlerAdapter which calls (handler core stats
   channel-handler-context message) for each received message.
 
+  To prevent Netty outbound buffer from filling up in the case of clients not 
+  reading ack messages, we close the channel when it becomes unwritable. Clients
+  should then be ready to reconnect if need be as they will receive some form
+  of exception in this case.
+
   Automatically handles channel closure, and handles exceptions thrown by the
   handler by logging an error and closing the channel."
   [core stats ^ChannelGroup channel-group handler]
@@ -61,6 +66,11 @@
     (channelActive [ctx]
       (.add channel-group (.channel ctx)))
 
+    (channelWritabilityChanged [^ChannelHandlerContext ctx]
+      (let [channel (.channel ctx)]
+        (when (not (.isWritable channel))
+          (.close channel))))
+    
     (channelRead [^ChannelHandlerContext ctx ^Object message]
       (try
         (handler @core stats ctx message)
@@ -93,6 +103,7 @@
     (.. ctx
       ; Actually handle request
       (writeAndFlush (handle core message))
+
       ; Record time from parse to write completion
       (addListener
         (reify ChannelFutureListener
--- a/src/riemann/transport/tcp.clj
+++ b/src/riemann/transport/tcp.clj
@@ -69,6 +69,8 @@
     (channelWritabilityChanged [^ChannelHandlerContext ctx]
       (let [channel (.channel ctx)]
         (when (not (.isWritable channel))
+          (warn "forcefully closing connection from " (.remoteAddress channel)
+                ". Client might be not reading acks fast enough or network is broken")
           (.close channel))))
     
     (channelRead [^ChannelHandlerContext ctx ^Object message]
--- a/test/riemann/transport_test.clj
+++ b/test/riemann/transport_test.clj
@@ -12,10 +12,13 @@
             [riemann.transport.sse :refer [sse-server]]
             [cheshire.core :as json]
             [riemann.client :as client]
+            [riemann.codec :as codec]
             [riemann.index :as index])
   (:import (java.net Socket
+                     SocketException
                      InetAddress)
-           (java.io IOException)))
+           (java.io IOException
+                    DataOutputStream)))
 
 (logging/init)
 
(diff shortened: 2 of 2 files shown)


aphyr:
Let's... go with it! I'm a little wary about folks getting disconnected in the wild so we'll just keep an eye out for reports.

aphyr:
Oh actually can you do me a favor @abailly and add a logging message when we do this with the client address, saying it might be a misbehaved client or a broken network?

eric:
Sorry to swoop in here (I haven't had a chance to fully understand the implications of this change), but is this going to also abruptly disconnect websocket clients that are listening on noisy queries?

abailly:
```
lein test riemann.transport-test
INFO [2016-01-10 08:19:59,505] clojure-agent-send-off-pool-0 - riemann.pubsub - PubSub shutting down.
INFO [2016-01-10 08:19:59,508] clojure-agent-send-off-pool-1 - riemann.transport.tcp - TCP server 127.0.0.1 15555 online
INFO [2016-01-10 08:19:59,508] main - riemann.core - Hyperspace core online
WARN [2016-01-10 08:20:03,230] defaultEventExecutorGroup-2-2 - riemann.transport.tcp - forcefully closing connection from  #<InetSocketAddress /127.0.0.1:57941> . Client might be not reading acks fast enough or network is broken
INFO [2016-01-10 08:20:03,396] main - riemann.core - Core stopping
INFO [2016-01-10 08:20:04,596] clojure-agent-send-off-pool-0 - riemann.transport.tcp - TCP server 127.0.0.1 15555 shut down
INFO [2016-01-10 08:20:04,596] main - riemann.core - Hyperspace core shut down
```

abailly:
@eric This change only affects `riemann.transport.tcp` which is the endpoint used by clients to push data streams. Websockets and SSE clients use `riemann.transport.sse` and `riemann.transport.websockets` which are not based on Netty, AFAIK.

eric:
Thanks for the details. This seems like a very clever solution to this problem. Nice job.

abailly:
Thanks @eric but honestly, I think this is more a workaround. I am a bit shocked that Netty does not offer a more straightforward way to handle this. I would expect that calling `writeAndFlush` or `write` when the underlying channel is unwritable would throw an exception, or would return a value like, say number of bytes written, or whatever to tell caller its operation can not complete.

I suggested earlier to simply not write ack when channel becomes unwritable but this is probably worse as client will not even get notified. We could imagine sending OOB data to signal the problem but then this would mean clients should handle OOB, and why would they do when they don't even handle ACKs?

eric:
As I think about this more, I wonder if it would be better to just stop reading if we can't write to give some sort of back-pressure to the client.

There have been times where I've done a query for `true` to get back all of the events in the index (and I have a lot of events in my index). I suspect that would be tripped by this code.

Maybe back-pressure like this would be a better idea:

``` clojure
   (channelWritabilityChanged [^ChannelHandlerContext ctx]
     (let [channel (.channel ctx)
           config  (.config channel)]
       (if (and (.isWritable channel) (not (.isAutoRead config)))
         (.setAutoRead config true)
         (.setAutoRead config false))))

```

abailly:
@eric That's much more subtle... How could I write a test for that?

eric:
You would have to do it with timeouts on the writes in your client. I agree that is more complicated to pull off.

pyr:
I'm inclined to go ahead and merge this for now.

pyr:
review done, thanks a lot @abailly

PR #645, opened 2016-01-15
Allow Cloudwatch to authenticate using IAM Instance Roles

iramello:
Currently, riemann.cloudwatch works with AWS keys. This commit enables riemann to authenticate using IAM Instance Roles.

### DIFF
--- a/src/riemann/cloudwatch.clj
+++ b/src/riemann/cloudwatch.clj
@@ -33,7 +33,7 @@
 
   :namespace   AWS CloudWatch namespace."
   [opts]
-  (let [opts (if (or (contains? opts :access-key) (contains? opts :access-key))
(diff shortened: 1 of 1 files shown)


src/riemann/cloudwatch.clj
```diff
@@ -31,10 +33,15 @@
 
   :namespace   AWS CloudWatch namespace."
   [opts]
-  (let [opts (merge {:access-key "aws-access-key"
+  (let [opts (if (or (contains? opts :access-key) (contains? opts :access-key))
```
pyr: You are using `or` on the same condition twice.

pyr:
Apart from the nitpick, I'd want to see a test case for your addition and then we're good to go.

PR #653, opened 2016-02-09
project: move to clojure 1.8

pyr:
I would love people to start giving this a shot.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -8,7 +8,7 @@
   :maintainer {:email "aphyr@aphyr.com"}
   :dependencies [
     [org.clojure/algo.generic "0.1.2"]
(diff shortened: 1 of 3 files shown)


AeroNotix:
Any expected breakages?

pyr:
@AeroNotix nope, but it's a big version bump and there are a lot of dependencies, so tests would be very appreciated.

AeroNotix:
I'll check it out on gizmos I use riemann with

pyr:
@AeroNotix there's a small warning for now since some symbols in riemann.index are now reserved in clojure 1.8 (namely: `update`) I'll adapt index later

PR #654, opened 2016-02-09
adding missing paren to fixed-time-window example

youngderekm:
adding missing paren, closing the the first argument to smap (the fn).


pyr:
sorry @youngderekm i committed a fixed in the meantime :-(

youngderekm:
hi @pyr - I think only a single paren should be added but your commit has two.  Can you double check?  I was using clojure-mode in emacs to double check the nesting level.

pyr:
@youngderekm you're right, the paren matching within markdown got me confused.

PR #656, opened 2016-02-18
fixed maintenance-mode function name in example

youngderekm:
The example defines `maintenance-mode` but then uses `maintenance-mode?`.  This change makes the example define `maintenance-mode?`.


jamtur01:
Thanks! Good catch!

PR #658, opened 2016-02-29
Use /bin/sh for tar package

pdericson:
The shell script doesn't appear to contain anything bash specific.


Anvil:
I would instead recommend to use bash arrays and make the script more robust. Using one variable for a list of items can be dangerous and/or can fail.

pdericson:
I don't need this anymore, I was working on a Docker image for riemann based on Alpine Linux but I didn't finish it.

PR #659, opened 2016-03-01
Expiring (by)

bergundy:
This is my take on limiting the resource consumption of (by) (https://github.com/riemann/riemann/pull/526).

I actually offer 2 different solutions, in 2 different commits.
The first commit provides a generic way of limiting the streams (by ttl / lru / whatever cache implementation you want - see: https://github.com/clojure/core.cache).
The second commit provides an expiry based solution and is not as generic.

I'd like to get some feedback on this, not sure which direction is better.

### DIFF
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -37,6 +37,7 @@
         clojure.tools.logging)
   (:require [riemann.folds :as folds]
             [riemann.index :as index]
+            [riemann.ttl-cache :refer [ttl-cache-factory]]
             riemann.client
             riemann.logging
             [clojure.set :as set])
@@ -1492,22 +1493,28 @@
   ; table is a reference which maps (field event) to a fork (or list of
   ; children).
   `(let [new-fork# (fn [] [~@children])]
-     (by-fn ~fields new-fork#)))
+     (by-fn ~fields new-fork# hash-map)))
 
-(defn by-fn [fields new-fork]
+(defmacro expiring-by
+  "Same as (by) but accepts ttl as an additional first argument which sets the lifespan of each fork"
+  [ttl fields & children]
+  `(let [new-fork# (fn [] [~@children])]
+     (by-fn ~fields new-fork# #(ttl-cache-factory {} :ttl (* ~ttl 1000)))))
+
+(defn by-fn [fields new-fork table-factory]
   (let [fields (flatten [fields])
         f (if (= 1 (count fields))
             ; Just use the first function given applied to the event
             (first fields)
             ; Return a vec of *each* function given, applied to the event
             (apply juxt fields))
-        table (atom {})]
-     (fn stream [event]
-       (let [fork-name (f event)
-             fork (if-let [fork (@table fork-name)]
-                    fork
-                    ((swap! table assoc fork-name (new-fork)) fork-name))]
-         (call-rescue event fork)))))
+        table (atom (table-factory))]
+    (fn stream [event]
+      (let [fork-name (f event)
+            fork (if-let [fork (get @table fork-name)]
+                  fork
+                  (get (swap! table assoc fork-name (new-fork)) fork-name))]
+        (call-rescue event fork)))))
 
 (defn changed
   "Passes on events only when (f event) differs from that of the previous
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -1493,28 +1493,45 @@
   ; table is a reference which maps (field event) to a fork (or list of
   ; children).
   `(let [new-fork# (fn [] [~@children])]
-     (by-fn ~fields new-fork# hash-map)))
+     (by-fn ~fields new-fork#)))
 
-(defmacro expiring-by
-  "Same as (by) but accepts ttl as an additional first argument which sets the lifespan of each fork"
-  [ttl fields & children]
-  `(let [new-fork# (fn [] [~@children])]
-     (by-fn ~fields new-fork# #(ttl-cache-factory {} :ttl (* ~ttl 1000)))))
-
-(defn by-fn [fields new-fork table-factory]
+(defn by-fn [fields new-fork]
   (let [fields (flatten [fields])
         f (if (= 1 (count fields))
             ; Just use the first function given applied to the event
             (first fields)
             ; Return a vec of *each* function given, applied to the event
             (apply juxt fields))
-        table (atom (table-factory))]
+        table (atom {})]
+     (fn stream [event]
+       (let [fork-name (f event)
+             fork (if-let [fork (@table fork-name)]
+                    fork
+                    ((swap! table assoc fork-name (new-fork)) fork-name))]
+         (call-rescue event fork)))))
+
+(defmacro expiring-by
+  "Same as (by), child streams created here will be garbage collected after their last seen event expires."
+  [fields & children]
+  `(let [new-fork# (fn [] [~@children])]
+     (expiring-by-fn ~fields new-fork#)))
+
+(defn expiring-by-fn [fields new-fork]
+  (let [fields (flatten [fields])
+        f (if (= 1 (count fields))
+            ; Just use the first function given applied to the event
+            (first fields)
+            ; Return a vec of *each* function given, applied to the event
+            (apply juxt fields))
+        table (atom (ttl-cache-factory {}))]
     (fn stream [event]
       (let [fork-name (f event)
+            expires (+ (:time event 0) (:ttl event 0))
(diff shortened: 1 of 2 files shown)


pyr:
Hi @bergundy!

Thanks for this, it's definitely a good improvement.
Here are my first thoughts:

- We are likely to extend by-fn to call new-fork with an additional argument, see #661
- I am wondering if an expiry-thread approach wouldn't be sufficient (you'd have to keep a list of by- streams though)

It would make the expiry opt-in, just like for events today.

bergundy:
Hi @pyr!
Thanks for the feedback.
I agree expiry should be opt-in (it is in both of my proposed solutions).
Maybe we can combine #661 and my solution.
I'm not sure about the expiry thread approach, using a priority-map for tracking expiration makes the operation pretty cheap to call on the main thread.

bergundy:
@aphyr, I'd really like to know what you think about this solution.
We've been running it in production for 2 weeks now and it works flawlessly

aphyr:
I'd like to see benchmarks for performance impact for narrow and wide fork cardinalities, and if it's acceptably fast and correct, recommend making expiry the default behavior with an option to use the old version where speed is more important.
As far as correctness, by fork expiry is tricky. Consider an index downstream of a by. A fork could drop any pattern of events, which means that you need to compute the expiry time of each incoming event (TTL + time) and use the maximum of those times as the new expiry time for the fork.
Another question: should by generate and dispatch an expired event just before expiring a fork? My sense is no: expiry is multithreaded which could lead to races where multiple levels of the stream generate and emit their own expired events.
I don't trust myself to code review this, haha. I tried, like, five approaches to by expiry and got it wrong each time. The best way to verify is probably careful generative testing.
On Mar 14, 2016 1:19 AM, Roey Berman <notifications@github.com> wrote:@aphyr, I'd really like to know what you think about this solution.
We've been running it in production for 2 weeks now and it works flawlessly

aphyr:
Oh, also look out for races between the garbage collection of old forks and the creation of new ones--you may need a barrier to ensure ordering. Or maybe not! Haven't thought carefully about it.

bergundy:
Hi @aphyr,

Thanks for the quick reply.

I'd be happy to provide some benchmarks but first I'd like to know that we're heading in the right direction.
Please note that I've provided 2 different solutions which approach the problem differently.
The [first](https://github.com/riemann/riemann/pull/659/commits/35018012292521f555a32638d4bfb397b8782d15) is very generic and simple and allows using different strategies (LRU / TTL / ..).
The [second](https://github.com/riemann/riemann/pull/659/commits/26267ccfa57d3a10a4e587c4e3b1f89e18242f9f) is more implicit and relies on the `:ttl` and `:time` of the event.
Using the first solution will give users more control over the amount of memory riemann uses (e.g: LRU) while the second provides a more seamless integration. 

I'm not sure expiry should be the default behaviour since it will probably break configurations using the non-expiring (by) mechanism.

> As far as correctness, by fork expiry is tricky. Consider an index downstream of a by. A fork could drop any pattern of events, which means that you need to compute the expiry time of each incoming event (TTL + time) and use the maximum of those times as the new expiry time for the fork.

This is addressed [here](https://github.com/riemann/riemann/pull/659/commits/26267ccfa57d3a10a4e587c4e3b1f89e18242f9f#diff-51b2d6e2c6d6f62d1e1a8aec145d84f5R1531).

> Another question: should by generate and dispatch an expired event just before expiring a fork? My sense is no: expiry is multithreaded which could lead to races where multiple levels of the stream generate and emit their own expired events.

I agree.

As for races and correctness, we might have to wrap the `get`, `set` sequence in a `dosync` block like so:

```clojure
        (dosync
          (let [fork (if-let [fork (get @table fork-name)]
                       {:stream (:stream fork) :expires (max expires (:expires fork))}
                       {:stream (new-fork) :expires expires})]
            (ref-set table assoc fork-name fork)))
```

I'm not really a clojure expert but from the reading I've done online it seems like this is needed.

aphyr:
LRU is definitely not correct as far as Riemann's event model is concerned, and you can't really say much about memory bounds in Riemann anyway. I would stick to TTL.
For races, I think careful use of an atom and maybe a lock should suffice. Maybe high-scale-lib's concurrenthashmap would be faster. Avoid dosync at all costs; perf will be terrible.

bergundy:
OK, so let's go with the :ttl and :time based solution.
I think we can't use a concurrenthashmap since we're actually manipulating 2 maps (one for the stream and one for the expiry).
I'm hoping to get around to writing the benchmarks this week.

bergundy:
I finally got around to conducting the benchmarks.

I ran into some gc issues while testing and I had to figure out exactly what happened (more on that later).

The riemann configuration I used is:

```clojure
(tcp-server {:host "127.0.0.1" :port 5555})
(instrumentation {:enabled? false})

(streams
  (if (= (System/getenv "GROUPER") "expiring-by")
    (do
      (prn "using (expiring-by)")
      (expiring-by :key bit-bucket))
    (do
      (prn "using (by)")
      (by :key bit-bucket))))
```

And the client code is:

```clojure
(ns riemann-expiring-by-bench.core
  (:require [riemann.client :as r])
  (:import (java.util.concurrent Executors))
  (:gen-class))

(defn rate [interval observable]
  (let [prev-state (atom 0)]
  (while true
    (Thread/sleep interval)
    (println (str (- @observable @prev-state) "/s"))
    (swap! prev-state (fn [_] @observable)))))

(defn -main
  [cardinality num-events-to-send]
  (with-open [c (r/tcp-client {:host "localhost"})]
    (let [ks (into [] (map str (range (Integer. cardinality))))
          num-events-to-send (atom (Integer. num-events-to-send))
          events-sent (atom 0)
          nthreads 16
          pool (Executors/newFixedThreadPool nthreads)
          base-event {:ttl 5 :metric 1}
          task (fn []
                 (let [chunk-size 1000]
                   (loop []
                     (when (>= (swap! num-events-to-send - chunk-size) 0)
                       @(r/send-events c (repeatedly chunk-size #(assoc base-event :key (rand-nth ks))))
                       (swap! events-sent + chunk-size)
                       (recur)
                     ))))]
      (doto
        (Thread. #(rate 1000 events-sent))
        (.setDaemon true)
        (.start))
      (time
        (let [futures (.invokeAll pool (replicate nthreads task))]
          (map #(.get %) futures)))
      (.shutdown pool))))
```

The results on my 16GB / 8 core Macbook Pro are:

grouper | cardinality | num items | total msecs
---------- | -------------- | ---------------- | -----------
(expiring-by) | 10000 | 10000000 | 109825
(by) | 10000 | 10000000 | 23387
(expiring-by) | 10000000 | 10000000 | 94205
(by) | 10000000 | 10000000 | 31880


(expiring-by) processing rate is inconsistent. You see variations from 18000 events/s to 188000 events/s.
This is not a problem with the regular (by) function where the rate can get up to 480000 events/s in the low cardinality set and gradually speeds up from ~200K events/s to ~400K events/s using the high cardinality set.

Running riemann with the default jvm configuration and 10M keys seems to cause problems in gc.
The allocated heap size keeps growing.

It starts out like this:
screen shot 2016-03-23 at 3 46 08 pm

And becomes this after 2 minutes:
screen shot 2016-03-23 at 3 45 36 pm

Adding the `-XX:MaxHeapFreeRatio=50` JVM option make the memory consumption more controllable but at a cost of more GC pauses which are noticable when you monitor the send rate.

screen shot 2016-03-23 at 3 49 11 pm

faxm0dem:
FWIW I wrote a [riemann plugin](/faxm0dem/riemann-mbeans) that streams jmx mbeans from the running jvm maybe that can help you debug even further

bergundy:
Thanks @faxm0dem,

I don't know if it'll give me any information that I can't get from yourkit but I might use this to monitor riemann in production.

bergundy:
@aphyr, your thoughts please?

aphyr:
Uh, this memory issue looks pretty painful and I don't think users would enjoy it?

bergundy:
@aphyr, keep in mind this benchmark tests the limits of expiring (by).
I think in real world scenarios users usually run much more complicated streams inside (by) than a simple bit-bucket.

How do you suggest we move forward?
I can provide more benchmarks with more complicated streams inside (by) if needed

bergundy:
@aphyr, bumping this in case you missed my last comment

pyr:
I think this would be best served as a separate fn, not changing the default by behavior.

pyr:
To expand on this @bergundy thanks a lot for all the work so far. It's a tough nut to crack and a very good start to making things better.

PR #661, opened 2016-03-07
streams: add by-builder and an additional arg to new-fork

pyr:
This allows building forked streams based on the fork name.

cc @Anvil


Anvil:
Hello.
I've integrated this code in my Riemann.
I can say it's working and it has simplified the code a lot comparing to the implementation I planned to write.

PR #663, opened 2016-03-17
Twilio REST API POST parameter names are updated.

metebalci:
Twilio REST API (https://www.twilio.com/docs/api/rest/sending-messages)
requires the form parameters named: From, To and Body; whereas
the current implementation sends from, to and body. This causes the
following errors returned from Twilio REST API:

A 'From' phone number is required.
A 'To' phone number is required.
Message body is required.


pyr:
Thanks @metebalci !

PR #667, opened 2016-03-22
slack: properly setup :fallback in bundled formatters

mfournier:
Adds a short plain-text summary to the slack attachments generated. This
is the text displayed by clients which don't support fancy formatting,
such as the slack IRC gateway.

### DIFF
--- a/src/riemann/slack.clj
+++ b/src/riemann/slack.clj
@@ -11,7 +11,7 @@
   (escape message {\< "&lt;" \> "&gt;" \& "&amp;"}))
 
 (defn slack-fallback
-  ""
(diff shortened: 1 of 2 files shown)


mfournier:
Prior to this patch, when simply using the example snippet for slack from the howto page, notifications from riemann print out "NO FALLBACK DEFINED" on IRC, instead of the event details, which is pretty useless...

mfournier:
I just bumped the commit to force travis-ci to run again, as the previous build failed because of a network issue.

mfournier:
@jamtur01,  @pyr, here's the reference & examples of the message format we're generating: https://api.slack.com/docs/attachments

pyr:
LGTM

PR #670, opened 2016-03-25
hipchat: remove threading as dissoc takes key sequnce as arguments

thenonameguy:


pyr:
thanks @thenonameguy !

PR #674, opened 2016-04-05
fix mistakes in credits

mfournier:


pyr:
sorry about this @mfournier !

PR #677, opened 2016-04-10
slack: fix tags handling

mfournier:
Previously the code was referring to the event's tags as `:tag` instead
of `:tags`, preventing them from getting displayed in the messages
generated by the default formatter.


mfournier:
I just amended the patchset with a test + bugfix for `slack/extended-formatter`.

pyr:
Thanks @mfournier

PR #679, opened 2016-04-11
Add batch forwarding for datadog

thearthur:
Make the datadog forwarder accept either single events or batches.
This greatly reduces the latency and increases throughput by at
least one order of magnatude.

Signed-off-by: Arthur Ulfeldt <arthur@ulfeldt.com>


pyr:
Thanks @thearthur

PR #681, opened 2016-05-02
Allow explicit config of Slack HTTP connection params

dbriones:
This PR allows users to configure the HTTP connection to Slack; the
params are passed directly to `clj-http`. This allows configuration of 
connection and socket timeouts, for example.


pyr:
Thanks @dbriones

Issue #686, opened 2016-05-11
Client might be not reading acks fast enough or network is broken


faxm0dem:
This is the error message I'm getting since `0.2.11` when issuing a query to the tcp-server having a reasonable-ish number of results (112). When removing the `(channelWritabilityChanged)` block from the code, everything runs smooth, I'm even able to query `true` which on my test server returns 18574 results.

Tested with `riemann-c-client` and `perl(Riemann::Client)`.

As discussed on IRC @abailly consider yourself poked

abailly:
I do :-)

faxm0dem:
just for the record

```
☠ time riemann-client query 'true' | grep -c Event
18574

real    0m0.963s
user    0m0.341s
sys     0m0.132s
☠ riemann-client query 'service =~ "%index size%"'
Event #0:
  time  = 1462957956 - Wed May 11 11:12:36 2016
  state = ok
  service = riemann index size
  host = ccosvms0046
  description = (null)
  ttl = 20.000000
  metric_sint64 = 18574
  metric_f = 18574.000000
```

abailly:
Quick question: does it also fails when using localhost interface? Just
asking because it would help in writing test case...

faxm0dem:
yeah, I still get the message

abailly:
Cool. Thanks and apologies for the gross simplification in client
handling....

faxm0dem:
please don't apologize for contributing!

PR #689, opened 2016-05-20
Don't close channel when buffer is overflowing

abailly:
This is a tentative fix to #686. When the channel becomes unwritable, instead of closing we now look at the response message to send: if it is a simple ACK message, we simply drop it. 

This behaviour is pretty simple and the rationale is the following:

* query responses can fill up buffer in case of slow clients, which is probably fine as long as query response do not grow too large to cause OOME, so let's keep writing those responses and let Netty's queues handle them
* when writing events' ACKs, if the buffer is filling up then this most probably means the client is not reading them properly: If it were slow, it would also probably be slow in sending events. Hence if it is not reading ACKs, don't send them back.

Comments welcomed


abailly:
The latest changes pushed simply remove closing channel when `writeabilityChanged` event is fired and revert previously proposed changes.

pyr:
This should be reopened or rebased with a single commit.

faxm0dem:
@abailly could you rebase this please - as reported by @dch this is still not merged to master

dch:
I ran into this issue last week, glad to find this fixes it. IMHO this would be worth rolling a 0.2.12 for, if I can help in any way to prepare that please let me know.

abailly:
I will do it later on today (CET).

abailly:
Done #724

PR #701, opened 2016-06-21
Add the ability to read SNS credentials from the default credential chain

hexedpackets:
This allows running Riemann on an EC2 instance which uses an IAM role for authenticating with SNS instead of explicitly creating and passing user credentials.


jamtur01:
Seems reasonable to me. Ping @pyr

pyr:
Thanks @hexedpackets!

PR #702, opened 2016-06-23
Correction in maintenance-mode function

pradeepchhetri:
From Riemann 0.2.8, we started representing fields as keywords instead of symbols and maintenance-mode documentation was out of date. This fixes that.


jamtur01:
Thanks! Merging.

PR #708, opened 2016-07-20
Fix fraction divisor in generating events example

guillaumebreton:
The fraction operation was using "events" as a divisor. "Events" is a
collection, so this patch replace it by the number of events, in order
to get the percentage of events.


pyr:
Thanks @guillaumebreton

PR #714, opened 2016-08-15
Telegram notification support

islander:
Implement sending notification to Telegram chats

### DIFF
--- a/src/riemann/telegram.clj
+++ b/src/riemann/telegram.clj
@@ -5,29 +5,44 @@
 
 (def ^:private api-url "https://api.telegram.org/bot%s/%s")
 
-(defn- format-message [ev]
+(defn- html-parse-mode []
+  "Formats html message."
+  (fn [e]
+    (str
+      "<strong>Host:</strong> " (or (:host e) "-") "\n"
+      "<strong>Service:</strong> " (or (:service e) "-") "\n"
+      "<strong>State:</strong> " (or (:state e) "-") "\n"
+      "<strong>Metric:</strong> " (or (:metric e) "-") "\n"
+      "<strong>Description:</strong> " (or (:description e) "-"))))
+
+(defn- markdown-parse-mode []
+  "Formats markdown message."
+  (fn [e]
+    (str
+      "*Host:* " (or (:host e) "-") "\n"
+      "*Service:* " (or (:service e) "-") "\n"
+      "*State:* " (or (:state e) "-") "\n"
+      "*Metric:* " (or (:metric e) "-") "\n"
+      "*Description:* " (or (:description e) "-"))))
+
+(defn- format-message [parse-mode event]
   "Formats a message, accepts a single
   event or a sequence of events."
   (join "\n\n"
         (map
-          (fn [e]
-            (str
-              "<strong>Host:</strong> " (or (:host e) "-") "\n"
-              "<strong>Service:</strong> " (or (:service e) "-") "\n"
-              "<strong>State:</strong> " (or (:state e) "-") "\n"
-              "<strong>Metric:</strong> " (or (:metric e) "-") "\n"
-              "<strong>Description:</strong> "
-                (escape (or (:description e) "-") {\< "&lt;", \> "&gt;", \& "&amp;"}))) 
-          (flatten [ev]))))
+          (if (re-matches #"(?i)html" parse-mode)
+            (html-parse-mode)
+            (markdown-parse-mode))
+            (flatten [event]))))
 
 (defn- post
   "POST to the Telegram API."
-  [token chat_id event]
+  [token chat_id event parse_mode]
   (client/post (format api-url token "sendMessage")
-               {:form-params {:chat_id chat_id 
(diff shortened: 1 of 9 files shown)


jamtur01:
Looks like some test failures in there.

src/riemann/telegram.clj
```diff
@@ -0,0 +1,48 @@
+(ns ^{:doc "Send events to Telegram"}
+  riemann.telegram
+  (:require [clj-http.client :as client]
+            [clojure.string :refer [join]]))
```
pradeepchhetri: You need to add `clojure.string/escape` here, that's why the test is failing.

src/riemann/telegram.clj
```diff
@@ -0,0 +1,48 @@
...
+              "<strong>Description:</strong> "
+                (escape (or (:description e) "-") {\< "&lt;", \> "&gt;", \& "&amp;"}))) 
+          (flatten [ev]))))
+
+(defn- post
+  "POST to the Telegram API."
+  [token chat_id event]
+  (client/post (format api-url token "sendMessage")
+               {:form-params {:chat_id chat_id 
+                              :parse_mode "HTML"
```
pradeepchhetri: Something good to have: `Markdown` support and exposing `parse_mode` as configurable parameter of the plugin.

src/riemann/telegram.clj
```diff
@@ -0,0 +1,48 @@
...
+
+  Usage:
+
+  (def token \"define_your_token\")
+  (def chat_id \"0123456\")
+
+  (streams
+    (rollup 5 3600 (telegram {:token token :chat_id chat_id})))"
+  [{:keys [token chat_id]}]
+  (fn [e] (post token chat_id e)))
```
pradeepchhetri: Looks like telegram have messages limit of 1 message per second. It will be nice if we can batch up events and call telegram api once every second (as default). Something for your reference as done in datadog plugin (https://github.com/riemann/riemann/blob/master/src/riemann/datadog.clj#L56-L59)

islander: If I understand correct, in datadog plugin batching events made outside of plugin (using native riemann's `batch` stream).

pradeepchhetri: yes you are right. you need to see if telegram api supports posting a list/vector of messages  in one HTTP call.

pradeepchhetri-nd:
@islander Also you need to update these files as well:

https://github.com/riemann/riemann/blob/master/project.clj
https://github.com/riemann/riemann/blob/master/src/riemann/config.clj

islander:
@pradeepchhetri-nd thank, i'll take a look later.

So, I just need to add `[telegram    :refer [telegram]]` in `:require` section of riemann/config.clj and `:telegram :telegram` to `:test-selectors` of project.clj?

P.S. Sorry for my mistakes and stupid questions, I'm new to clojure and riemann development =)

jamtur01:
@islander Yes.

src/riemann/telegram.clj
islander: in Telegram API docs I don't see such method, so I just map all events in one big message.

pradeepchhetri-nd: Then we can go ahead with this.  Thanks for searching about this.

islander: I think you're right, we need some batch method. Now I get this error from Telegram API when I try to send a lot of  events in one message: `{"ok":false,"error_code":400,"description":"Message is too long"}`

jamtur01:
LGTM - excluding the configurable formatting.

pradeepchhetri-nd:
@jamtur01 @islander I will try this on my setup today and how the testing goes. Thanks.

jamtur01:
@pradeepchhetri I am not overly keen on the `sleep`....

pradeepchhetri:
@jamtur01 @islander I think we can do something better than `sleep` Will get back to you soon.

islander:
@jamtur01 by external formating you mean switching between HTML and Markdown? I'll try to add this soon. And what other options we have to replace `sleep`?

islander:
According to formatting: sometimes I send, for example, syslog message in event description, so telegram markdown parser could interpret all special symbols in message ([,],*,_ and so on) and formatting become weird. So, should I make escaping inside format-message function, or it should be done outside of it (in client, or somewhere in riemann streams)? It's the reason i switched to HTML parse_mode (easier to escape :))

example of such message:
```
Description: WARNING[817][C-00036738]: app_dial.c:2437 in dial_exec_full: Unable to create channel of type 'SIP' (cause 20 - Subscriber absent)
```

becomes something like this:

Description: WARNING817C-00036738: app _dial.c:2437 in dial_ exec_full: Unable to create channel of type 'SIP' (cause 20 - Subscriber absent)

pradeepchhetri:
@islander I send a pull request to your fork's `telegram` branch. I tested the plugin and works perfectly fine.

islander:
I've added `parse_mode` option.

pyr:
thanks @islander

PR #731, opened 2016-10-18
Influxdb dynamic db rp

liuweichu:
This PR implements a feature that is described in issue 730 that by setting `db-fn` and `rp-fn`, `influxdb` can automatically determine which database/retention policy to forward.

The code is tested with our own deployment. It works, but I don't have time to test it with more cases yet.

The changes is kind of IO related, I checked `influxdb_test.clj` but don't have an idea how to add test cases for this feature.

This is the first time I made a PR to Riemann. Please kindly add comments here if I missed something


pyr:
superseded by #741

PR #735, opened 2016-10-25
fix websocket listening address

### LINKED ISSUE #721: ws-server binds to all interfaces/IP addresses even if config requires otherwise
This seems to be present in master, even though we switched `aleph` for `httpkit`. The original message is here https://github.com/riemann/riemann/issues/156 but if there is no fix possible it might make more sense to have `ws-server` throw an error if a `:host ...` is passed in as config.

mcorbin:
fix #721.
the :host option for ws-server was ignored. It is because http-kit uses the :ip keyword for the host.

### DIFF
--- a/src/riemann/transport/websockets.clj
+++ b/src/riemann/transport/websockets.clj
@@ -183,7 +183,7 @@
           (locking this
             (when-not @server
               (reset! server (http/run-server (ws-handler core stats)
-                                              {:host host
+                                              {:ip host
                                                :port port}))
               (info "Websockets server" host port "online"))))
 


dch:
```shell
# sockstat -46l |grep riem
riemann  java       25905 18 tcp4   127.0.0.1:5557        *:*
riemann  java       25905 82 tcp4   127.0.0.1:5556        *:*
riemann  java       25905 170 tcp4  127.0.0.1:5555        *:*
```

:+1:  working like a boss on FreeBSD. thanks!

aphyr:
Thanks @mcorbin!

Issue #739, opened 2016-11-01
[Feature request] Improve influxdb support


mcorbin:
Riemann should support latest influxdb options, like consistency level, better time precision...
We can use a clojure wrapper of [the officiel java client](https://github.com/influxdata/influxdb-java).
We should probably drop the influx 0.8 support.

PR #741, opened 2016-11-03
[Need review] Refactoring influxdb

mcorbin:
#739 
This is a major refactoring of the InfluxDB code.  The goal is to support all influxdb  options. I have wrapped the officiel influxdb java client. I await your opinions on this (is this change a good idea, possibles improvements in the code...). Normally, the change is transparent for the user (same options).

## Breaking change : 

- this code does not support InfluxDB < 0.9 (like the official java client). Is it acceptable ?

## Improvements :
- Support InfluxDB consistency levels.
- Support differents time units (micros, millis and seconds, automatic conversion) with a :precision keyword on the event.

The :insecure connection mode is actually untested. Otherwise, all tests are passing.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -20,7 +20,7 @@
     [ch.qos.logback/logback-classic "1.1.3"]
     [com.github.juise/logstash-logback-layout "1.0"]
     [net.logstash.logback/logstash-logback-encoder "4.5"]
-
+    [org.influxdb/influxdb-java "2.4"]
     [com.cemerick/pomegranate "0.3.0"
      :exclusions [org.codehaus.plexus/plexus-utils]]
     ; for pomegranate
--- a/src/riemann/influxdb.clj
+++ b/src/riemann/influxdb.clj
@@ -1,47 +1,36 @@
 (ns riemann.influxdb
-  "Forwards events to InfluxDB. Supports both 0.8 and 0.9 APIs."
+  "Forwards events to InfluxDB. Supports InfluxDB 0.9 or Higher"
   (:require
-    [capacitor.core :as capacitor]
-    [cheshire.core :as json]
-    [clj-http.client :as http]
-    [clojure.set :as set]
-    [clojure.string :as str]
-    [riemann.common :refer [unix-to-iso8601]]))
+    [clojure.set :as set])
+  (:import
+   (java.util.concurrent TimeUnit)
+   (javax.net.ssl SSLContext X509TrustManager HostnameVerifier)
+   (java.security SecureRandom)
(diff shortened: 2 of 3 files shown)


mcorbin:
There was a bug on the insecure https mode. I fixed it, and tested with a self signed certificate, it is working.

mcorbin:
Fix #742 (:tag-fields can be in the event map).

mcorbin:
I have updated the PR.

By default, (influxdb) will returns a stream with the actual behavior. Default options are :

```clojure
{:db "riemann"
   :scheme "http"
   :version :0.9
   :host "localhost"
   :username "root"
   :precision :seconds
   :port 8086
   :tags {}
   :tag-fields #{:host}
   :consistency "ONE"
   :timeout 5000
   :insecure false}
```
The `retention` option can also be set (no default value).
The `:precision` and `:tag-fields` keys can be set per-event. Possibles values for `precision` are `:seconds`, `:milliseconds` and `:microseconds`.

If the `:version` passed to `ìnfluxdb` is `:new-stream`, it will returns a new stream. This stream expects events like :

```clojure
{:time 1428366765
 :tags {:foo "bar"
           :bar "baz"}
 :precision :seconds
 :db "riemann_test"
 :consistency "ALL"
 :retention "autogen"
 :measurement "measurement"
 :fields {:alice "bob"}}
```
`:tags` are influxdb tags, `fields` are influxdb fields. `consistency`, `db`, `retention`, and `:precision` can be set per event (override the global stream settings). `:tag-fields` is not used.

mcorbin:
I have updated the PR (minor changes).

Also, the `lineprotocol-encode-9` fn is removed, because the java influxdb client [escape chars](https://github.com/influxdata/influxdb-java/blob/master/src/main/java/org/influxdb/dto/Point.java#L31).

mcorbin:
I am currently writing ** a lots** more tests for this PR (including integration tests using `with-mock`) to be sure i didn't miss anything.

@jamtur01 @pyr what do you think about the current event format described in my previous comment for the new influxdb stream?

pyr:
@mcorbin I would want doc updates to follow and a message to go out to the mailing list. Apart from that, i'm fine with the changes.

ecks:
Hello,

Does this change work with the influxdb 1.2.0?

mcorbin:
I wrap the official java client (https://github.com/influxdata/influxdb-java) so this PR support Influx 0.9 and higher.

mcorbin:
@pyr I did a PR for the doc but we cannot merge it now.

Issue #743, opened 2016-11-23
[Feature request] keep streams state between reloads


mcorbin:
Actually, streams like fixed-time-window, throttle, rollup, coalesce... lose their states between reloads. For example, throttle lose his counter and its time when reloading.
I think we can modify these streams to not lose their states. I was thinking about something to do that. For example, let's take the throttle stream :

`(throttle 1 60 #(info %))` will have the actual behavior (lose state between reloads). 
`(throttle 1 60 :my-stream-name #(info %))` will keep the stream state between reloads. The third parameter (**:my-stream-name**) is an identifier.

In src/riemann/streams.clj, i defined an atom :

```clojure
(def stream-state (atom {}))
```

This atom contains the state for each stream. 

Throttle use part-time-simple to manage its time window. If i want to keep the part-time-simple state, i need to keep the "[state](https://github.com/riemann/riemann/blob/master/src/riemann/streams.clj#L589)" atom (and maybe the "anchor" variable too).

The actual part-time-simple definition can be replaced by

```clojure
  ([dt reset add finish]
   (part-time-simple dt reset add (fn [state event]) finish nil))
  ([dt reset add side-effects finish]
   (part-time-simple dt reset add side-effects finish nil))
  ([dt reset add side-effects finish stream-name]
```

There is a new parameter (stream-name), which is in my case **:my-stream-name**
Now, at the beginning of part-time-simple, i :

- check if the stream-name (:my-stream-name) key is present in my **stream-state** atom
- If yes, i initialize the state variable with the value associated with the stream-name key, and i schedule (with once!) the next part-time-simple tick if the :scheduled key is present in the state (that means that riemann reloads before the next part-time-simple tick).
- If no, i initialize the state.
- I assoc the new state variable to stream-name key in **stream-state** (so stream-state = {:my-stream-name <state>}
- That it, the rest of the stream does not change.

I quickly tested this in a branch and it *seems* to work, state is not lost between reloads. But before going further, i need your opinion :

- Is losing the streams states between reloads is a problem for you ?
- What do you think about my way of doing it ?
- Do you have other ideas to keep streams states between reloads ?

I hope my explanations are clear ;)

faxm0dem:
I would love that feature merged. Maybe people rely on the current behaviour - we could make the change optional

Ben-M:
I'd like to have this! There are some use cases that our users have asked for that we've had to push back on because we are worried about losing state too frequently.

faxm0dem:
[...]

PR #744, opened 2016-11-28
Keep coalesce state between reloads

mcorbin:
This PR allows the coalesce stream to keep state between reloads ([#743]([https://github.com/riemann/riemann/issues/743))

### DIFF
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -1164,6 +1164,44 @@
             [ok expired] (swap! past update)]
         (child event (concat expired (vals ok)))))))
 
+(def stream-state (atom {}))
+(def next-stream-state (atom {}))
+
+(defn set-next-stream-state!
+  "Set the `next-stream-name` key in stream-state to `state`
+  `stream-name` must be a keyword"
+  [stream-name state]
+  (swap! next-stream-state assoc stream-name state))
+
+(defn reset-stream-states!
+  "Reset the `stream-state` and `next-stream-state` atoms to `{}`"
+  []
+  (reset! stream-state {})
+  (reset! next-stream-state {}))
+
+(defn stream-state-transition!
+  "Copy the `next-stream-state` to `stream-state`
+  reset the `next-stream-state` value to `{}`"
+  []
+  (reset! stream-state @next-stream-state)
+  (reset! next-stream-state {}))
+
+(defn get-or-create-stream-state
+  "Returns the value associated to the `stream-name` key in `stream-state` or `default-value` if the key does not exists.
+  Update the `next-stream-state` atom."
+  [stream-name default-value]
+  (let [state (get @stream-state stream-name default-value)]
+    (when stream-name (set-next-stream-state! stream-name state))
+    state))
+
+(defn extract-coalesce-args
+  "returns a map containing the coalesce args."
+  [dt children]
+  (cond
+    (number? dt) {:dt dt :children children}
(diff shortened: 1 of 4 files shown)


jamtur01:
LGTM but I think @pyr @aphyr are pretty placed to assess this than me.

src/riemann/streams.clj
```diff
@@ -1179,10 +1208,22 @@
...
+                   (keyword? dt) (cons stream-name children)
+                   true (remove nil? (conj children dt stream-name)))
+        stream-name (if (keyword? dt)
+                      dt
+                      (if (keyword? stream-name)
+                        stream-name
+                        nil))
         dt (if (number? dt) dt 1)
-        chm (java.util.concurrent.ConcurrentHashMap.)
+        chm (if-let [state (get-stream-state stream-name)]
```
pyr: `get-stream-state` should handle creation and updating the map.

```clojure
(defn get-or-create-state
     [stream-name]
     (get
         (update stream-state stream-name #(or % (java.util.ConcurrentHashMap.)))
         stream-name))
```

src/riemann/streams.clj
```diff
@@ -1179,10 +1208,22 @@
 
   (by [:foo :bar]
     (coalesce 10 prn))"
-  [& [dt & children]]
-  (let [children (if (number? dt) children (cons dt children))
+  [& [dt stream-name & children]]
+  (let [children (cond
```
pyr: As per the conversation comment, This should be extracted in its own function. For compatibility reasons, i'd keep looking for a number, but instead of testing the second argument, I'd handle the first argument being a map as well.

pyr:
Hi @mccorbin,

Thanks a lot for the PR. Very keen on moving this forward. I feel that we might be at a point where coalesce tries to do too much with its arguments.

I would argue for splitting the argument extraction in its own function.

```clojure
(let [{:keys [dt stream-name children]} (extract-coalesce-args args)]
      ...)
```

I would also change the behavior a bit to only look at the first element and apply the following heuristic:

- If it's a number, it's dt
- If it's a map, look for `:stream-name` and `:dt` keys in the map
- Otherwise, conj the first arg in the children list

src/riemann/streams.clj
mcorbin: I have created a new "get-or-create-stream-state" fn. The first arg is the stream name and the second arg the default value if the state does not already exists. I want to reuse this fn in the future on others streams (part-time-simple for example). What do you think of it ?

mcorbin:
Hi @pyr and thank you for your review.
I have updated my PR based on your comments.

pyr:
thanks @mcorbin

pyr:
@mcorbin any reason you opted against using defonce for coalesce's storage? this would avoid having to do accounting between `stream-state` and `next-stream-state`

mcorbin:
@pyr Do you mean using defonce for `the stream-state` variable ?
I use `stream-state` and `next-stream-state` to clean old/unused states.

Example : if i have  `(coalesce {:stream-name :foo} #(info %))` and start riemann, `next-stream-state` will contains the :foo state. If i rename the stream to `(coalesce {:stream-name :bar} #(info %))` and reloads, i will have `stream-state` with the :foo state (but unused), and `next-stream-state` with the :bar state. If i reload again, the :foo state will be cleaned.

I don't know how to use defonce for this use case, but i can do another PR is you have a solution.

PR #748, opened 2016-12-12
coalesce: simplify coalesce state storage

pyr:
Here's a proposed changed to how coalesce would store state.
This does away with the two atoms, and uses defonce to have a single persistent state atom.
For the odd-case where you might be tempted to change/remove a coalesce, a utility function is provided.


pyr:
cc @mcorbin for review.

mcorbin:
LGTM ;)

PR #749, opened 2016-12-14
Slack : cast event :tags into vector

mcorbin:
With this conf :

```clojure
(streams
 (tag "foo"
   slacker))
```

I have in slack :

```
Host:   host
Service:   riemann server tcp 0.0.0.0:5555 in latency 0.5
State:   ok
Description:   -
Metric:   -
Tags:   clojure.lang.LazySeq@18ce5
```
It is because the `tag` stream produces a LazySeq in :tags. This PR cast the :tags value into vector in the slack formatters.


jamtur01:
Thanks!

PR #750, opened 2016-12-17
Keep part-time-simple state between reloads

mcorbin:
In the same way as the coalesce stream, store the part-time-simple state in the `stream-state` atom. The `dt` parameter for streams like `batch`, `throttle`... can be a map containing the stream name and the dt value.

Can cause behavior like this one if `expire-stream-state` is not called :

- config with ```(batch 3 {:dt 400 :stream-name :foo} #(info %)```
- send 2 events to Riemann
- config with ```(batch 30 {:dt 15 :stream-name :foo} #(info %)```, reload
- Send 2 events to Riemann
- You need to wait the end of the previous tick (400 seconds) before having the desired behavior.

### DIFF
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -49,6 +49,36 @@
   "When an exception is caught, it's converted to an event and sent here."
   nil)
 
+(defonce stream-state (atom {}))
+
+(defn set-stream-state!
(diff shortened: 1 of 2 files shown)


mcorbin:
Not working because of [#758](https://github.com/riemann/riemann/issues/758)

PR #751, opened 2016-12-19
Clarify throttle doc-string

Ben-M:
It wasn't clear to me how throttle would handle vectors of events (for instance the output from moving-time-window). I've tried to clarify the doc-string.


jamtur01:
Thanks!

PR #752, opened 2016-12-19
Clarify splitting into multiple files.

Ben-M:
I had trouble working out how to split my config into multiple files. I've tried to clarify.

### DIFF
--- a/howto.html
+++ b/howto.html
@@ -165,14 +165,16 @@ sudo service riemann reload
   </div>
 </div>
 
-<h3>Including functions from multiple sources</h3>
+<h3>Including functions from multiple files</h3>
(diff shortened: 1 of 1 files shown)


jamtur01:
Not sure we quite yet want to remove the `include` docs - I suspect there's folks with older versions. But I'm open to persuasion or making it clearer they are deprecated.

Ben-M:
I'm sure there are people with older versions. However, do you think there are people with older versions who don't currently use `include` and who will start using it after reading the HowTo? If yes, then I'll think of other options.

jamtur01:
Meh. Probably right. :) 

Thanks!

PR #756, opened 2017-01-04
Add "riemann" tag to Instrumented transports and services

mcorbin:
Actually, core instrumentation produces events (rate and latency) with the "riemann" tag, but transports and services instrumentation do not.
This PR adds the "riemann" tag to instrumented transports and services.


jamtur01:
Thanks!

PR #762, opened 2017-01-19
Twilio : refactor options

### LINKED ISSUE #761: Accept more Twilio params and do not force 'from'
The twilio plugin seems to force the `from` param here https://github.com/riemann/riemann/blob/0.2.12/src/riemann/twilio.clj#L75 

The problem is when you use a dedicated `MessagingServiceSid` then you are require to send it instead of `from` as explained here : https://www.twilio.com/docs/api/rest/sending-messages#post-parameters-conditional 

Since `from` is forced in the twilio plugin the `MessagingServiceSid` is not used which is problematic.

I propose to stop forcing the `from` argument in the Twilio plugin and to let users configure it manually in their own Riemann config. This way only Twilio's credentials (`account` and `service-key`) are mandatory. The rest is up to the user.

What do you think ?

:)

mcorbin:
Fix #761

Allows more Twilio options and do not force a default :from if :messaging-service-sid is used.

Should be backward compatible with the current twilio stream. I used the twilio api doc [here](https://www.twilio.com/docs/api/rest/sending-messages#post-parameters-conditional) for the new options.

I have not tested it using twilio because i don't have an account, so i think it should be carefully review.

### DIFF
--- a/src/riemann/twilio.clj
+++ b/src/riemann/twilio.clj
@@ -6,20 +6,37 @@
 (def ^:private messages-url
   "https://api.twilio.com/2010-04-01/Accounts/%s/Messages.json")
 
+(defn add-key-body
+  "Add the `opts-key` value from `msg-opts` (if exists) into the `body-key` in `result`.
+  Returns result."
+  [opts-key body-key msg-opts result]
+  (if-let [value (get msg-opts opts-key)]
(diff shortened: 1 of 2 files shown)


jamtur01:
Thanks!

PR #765, opened 2017-01-25
Fix some tests for time micro

mcorbin:
I modified 2 tests in this PR :

- in core_test.clj : the (event) fn in riemann.common returns a new event but cast the :time field in long, example :

```clojure
(event {:time 1.1})
#riemann.codec.Event{:host nil, :service nil, :state nil, :description nil, :metric nil, :tags nil, :time 1, :ttl nil}
```

But now Riemann use double for the :time field. So the test fail. I modified it to cast the event :time field in double, and then transform the result into a set.

- transport_test.clj : same problem, (event) cast the event :time in long.

With this PR, all Riemann tests are passing. Maybe we should modify riemann.common/event in the future to not cast :time into long.

### DIFF
--- a/test/riemann/core_test.clj
+++ b/test/riemann/core_test.clj
@@ -239,11 +239,10 @@
                           (query client)
                           deref
                           set)]
-               (is (= r
-                      #{(event {:metric 2, :time 3})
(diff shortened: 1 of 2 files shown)


jamtur01:
@pyr ?

pyr:
All good, thanks @mcorbin

PR #772, opened 2017-02-14
Fix pagerduty documentation (:service-key parameter)

mcorbin:
Fix the Pagerduty documentation (:service-key is needed, cf [the api](http://riemann.io/api/riemann.pagerduty.html#var-pagerduty)).
I also added a link to the Riemann pagerduty api.


jamtur01:
Thanks!

PR #773, opened 2017-02-14
Add an `:options` parameter to the pagerduty stream

mcorbin:
This PR add an optional `:options` parameter to the pagerduty stream. It can be used to add more http options (like proxy) to the http client.
I also added a test file and wrote some tests for pagerduty.


jamtur01:
Nice! Thanks for adding these.

Issue #767, opened 2017-02-08
[Feature Request] Add kafka as input option


liuweichu:
Since producing events to kafka is added as output option, it will be nice to have kafka also as an input option.

boernd:
I would also welcome a kafka input ootb. You could address topics like availabilty or sharding with this kind of setup.

liuweichu:
Close this issue as PR get merged to master. Thanks @boernd

PR #781, opened 2017-02-20
add kafka as input option

boernd:
Addresses issue #767 

I ported a lot of code from https://github.com/pyr/riemann-kafka. As I'm not that familiar with the riemann internals, I'm not sure why the `clojure.lang.ILookup` implementation in the `kafka-consumer` function is needed (or maybe it's not anymore)? 

Tested with kafka 0.10.0.1

### DIFF
--- a/src/riemann/config.clj
+++ b/src/riemann/config.clj
@@ -18,7 +18,7 @@
                      [hipchat     :refer [hipchat]]
                      [index       :as index]
                      [influxdb    :refer [influxdb]]
-                     [kafka       :refer [kafka]]
+                     [kafka       :as kafka :refer [kafka]]
                      [kairosdb    :refer [kairosdb]]
                      [keenio      :refer [keenio]]
                      [librato     :refer [librato-metrics]]
@@ -173,6 +173,31 @@
   (sse-server {:port 5556})"
   [& opts]
   (service! (sse/sse-server (kwargs-or-map opts))))
+
+(defn kafka-consumer
+  "Add a new kafka consumer with opts to the default core.
+
+  (kafka-consumer {:consumer.config {:bootstrap.servers \"localhost:9092\"
+                                     :group.id \"riemann\"}
+                   :topics [\"riemann\"]})
+ 
+  Options:
+   
+  For a full list of :consumer.config options see the kafka consumer docs.
+  NOTE: The :enable.auto.commit option is ignored and defaults to true.
+
+  :consumer.config      Consumer configuration 
+    :bootstrap.servers  Bootstrap configuration, default is \"localhost:9092\"
+    :group.id           Consumer group id, default is \"riemann\"
+  :topics               Topics to consume from, default is [\"riemann\"]
+  :key.deserializer     Key deserializer function, defaults to the 
+                        keyword-deserializer.
+  :value.deserializer   Value deserializer function, defaults to 
+                        json-deserializer.
+  :poll.timeout.ms      Polling timeout, default is 100."
+
+  [& opts]
+  (service! (kafka/kafka-consumer (kwargs-or-map opts))))
 
 (defn streams
   "Add any number of streams to the default core."
--- a/src/riemann/kafka.clj
+++ b/src/riemann/kafka.clj
@@ -1,6 +1,11 @@
 (ns riemann.kafka
-  "Forwards events to Kafka."
-  (:require [kinsky.client :as client]))
+  "Receives events from and forwards events to Kafka."
+  (:require [kinsky.client :as client]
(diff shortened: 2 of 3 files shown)


src/riemann/kafka.clj
```diff
@@ -36,3 +41,70 @@
...
+  [opts]
+  (let [running? (atom true)
+        core     (atom nil)]
+    (reify
+      clojure.lang.ILookup
+      (valAt [this k not-found]
+        (or (.valAt this k) not-found))
+      (valAt [this k]
+        (info "Looking up: " k)
+        (if (= (name k) "opts") opts))
```
liuweichu: [nitpicks] here can be `when`

liuweichu:
Thank you @boernd very much for working on my Issue!

Here is one of our feature request: In kinsky a client supports subscribing to multiple topics (you just run `client/subscribe!` for several times.)

It will be very nice to use one client to subscribe multiple topics instead of spinning off many kafka clients.

liuweichu:
From our experience running `riemann-kafka` in production for almost two years, there is usually more than 1 topic you need to ask Riemann to subscribe to

src/riemann/kafka.clj
```diff
@@ -36,3 +41,70 @@
...
+(defn start-kafka-thread
+  "Start a kafka thread which will pop messages off the queue as long
+  as running? is true"
+  [running? core opts]
+  (let [opts (merge {:bootstrap.servers "localhost:9092"
+                     :topic "riemann"
+                     :key.deserializer client/keyword-deserializer
+                     :value.deserializer json-deserializer}
+                    opts)
+        consumer (client/consumer (dissoc opts :topic :key.deserializer :value.deserializer)
```
liuweichu: I think `kafka-consumer` function should accept two seperated opts (instead of mixing them up), one is `kafka-consumer` opts, another is the options that will be sent as-is to kinsky `client/consumer`.

Otherwise when more consumer options are added, here will be a mess.

src/riemann/kafka.clj
```diff
@@ -36,3 +41,70 @@
...
+        topic (:topic opts)]
+    (future
+      (try
+        (client/subscribe! consumer topic)
+        (while running?
+          (let [msgs (client/poll! consumer 100)
+                records (get-in msgs [:by-topic topic])]
+            (doseq [record records]
+              (let [event (post-load-event (get record :value))]
+                (stream! @core event)))))
```
liuweichu: If `:enable.auto.commit` is set to false(by default it is `true` though), there will be no way for user to commit manually.

I think this should be documented as Caveat.
(Or even nicer, when `:enable.auto.commit false` is detected, provide user a way to commit, or commit per polling for them)

src/riemann/kafka.clj
```diff
@@ -36,3 +41,70 @@
...
+                    opts)
+        consumer (client/consumer (dissoc opts :topic :key.deserializer :value.deserializer)
+                                  (:key.deserializer opts)
+                                  (:value.deserializer opts))
+        topic (:topic opts)]
+    (future
+      (try
+        (client/subscribe! consumer topic)
+        (while running?
+          (let [msgs (client/poll! consumer 100)
```
liuweichu: [nitpickings] Rather than hardcoded value, will be nice to be a configurable option with 100 as default

src/riemann/kafka.clj
```diff
@@ -36,3 +41,70 @@
...
+        (if (= (name k) "opts") opts))
+      ServiceEquiv
+      (equiv? [this other]
+        (= opts (:opts other)))
+      Service
+      (conflict? [this other]
+        (= opts (:opts other)))
+      (start! [this]
+        (info "Starting kafka consumer for topic: "
+              (:topic opts))
```
liuweichu: Should have a fallback: `(:topic(s) opts "riemann")`

src/riemann/config.clj
```diff
@@ -18,7 +18,7 @@
                      [hipchat     :refer [hipchat]]
                      [index       :as index]
                      [influxdb    :refer [influxdb]]
-                     [kafka       :refer [kafka]]
+                     [kafka       :as kafka :refer [kafka]]
```
liuweichu: Seems there's no need to do the `:as` since it will be overridden by the `:refer`.

boernd:
Thanks @liuweichu for reviewing, I addressed (most) of your comments.

liuweichu:
@boernd Thank you very much for addressing my opinions!

We are really excited to see this to be merged to master

Ben-M:
@jamtur01 anything missing to get this merged?

PR #784, opened 2017-03-14
Add pagerduty `:options` key doc and fix jekyll command

mcorbin:
Add documentation for the `:options` key in the pagerduty stream (to pass http options). Also fix the jekyll command to run it.

The `:options` key for the pagerduty stream will be available in the next Riemann release.


jamtur01:
LGTM. Thanks!

PR #785, opened 2017-03-17
Document inject! on specific streams

mcorbin:
Document how to use inject! to inject events on specific streams.

### DIFF
--- a/howto.html
+++ b/howto.html
@@ -782,6 +782,8 @@ set :bind, "1.2.3.4"
     like rate, rollup, etc will take effect atomically as the virtual clock
     advances.</p>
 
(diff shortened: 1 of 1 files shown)


howto.html
```diff
@@ -782,6 +782,8 @@ <h1 class="twelvecol">A <span class="focus">guide</span> to working with Riemann
     like rate, rollup, etc will take effect atomically as the virtual clock
     advances.</p>
 
+    <p>You can also inject events on specific streams using <code>(inject! [stream] events)</code>.
```
jamtur01: Should `on` be `into` here?

mcorbin: Fixed, thanks :D

jamtur01:
Thanks mate!

Issue #788, opened 2017-03-20
provide streams like runs but on duration


mcorbin:
I think it may be useful to have streams like `(runs)` but on duration instead of moving-event-window.

For example, i want to be alerted if :metric is > to a threshold (600 for example) during 5 minutes. It would be nice to have :

```clojure
(above 600 300 children)
```

If an event arrives at time 0 with :metric = 700, and a second event arrives at time 301 with :metric = 750, above passes on the last event.

Another example with a test case : 

```clojure
(test-stream (above 10 5) [{:time 0
                            :metric 2}  ;; metric < 10
                           {:time 6
                            :metric 11} ;; metric > 10, valid event
                           {:time 12 ;; metric > 10, 12 - 6 > 5 => send to children
                            :metric 11}] [{:time 12
                                           :metric 11}])
```

More generally, we could have a stream called cond-dt taking a function of events, a duration, and children, and use it to creates for example the above stream :

```clojure
(defn above
  [threshold dt & children]
  (apply cond-dt (fn [event] (> (:metric event) threshold)) dt children))
```

I am working on it actually (i will do a PR tomorrow), do you think it can be useful to integrate it into Riemann ?

mcorbin:
cf #789 and https://github.com/mcorbin/riemann-cond-dt-plugin ;)

PR #789, opened 2017-03-21
Add cond-dt, above, below, between, outside streams

mcorbin:
cf #788 

This PR adds a few new streams. 

## above example

"Takes a number `threshold` and a time period in seconds `dt`.
If the condition `(> (:metric event) threshold)` is valid for all events received during at least the period `dt`, valid events received after the `dt` period will be passed on until an invalid event arrives.
`:metric` should not be nil (it will produce exceptions)."

Let's begin with an example (a test case) :

```clojure
(test-stream (above 10 5) [{:time 0
                            :metric 11}
                           {:time 7
                            :metric 4}
                           {:time 9
                            :metric 11}
                           {:time 15
                            :metric 12}] [{:time 15
                                           :metric 12}])
```
In this test, the threshold is 10, dt is 5.
- At time 0, :metric is 11 (so > 10, event is valid)
- At time 7, :metric is 4 (< 10, invalid)
- At time 9, :metric is 11 (valid)
- At time 15, :metric is 12 (valid). During 6 seconds (15 - 9), :metric was > to threshold => above passes on this event.

Each new valid event will be forwarded to children, until an invalid event arrives.

## below, between, outside

below check if :metric is < to a threshold during dt seconds, between if :metric is inside ]low, high[ during dt seconds, outside if :metric is outside [low, high] during dt seconds.

## cond-dt

All these streams used cond-dt under the hood. cond-dt takes a function of events, a duration, and children. For example, the above stream is just :

```clojure
(apply cond-dt (fn [event] (> (:metric event) threshold)) dt children)
```
cond-dt can be used to creates new streams by users.

## limitations

Like in all Riemann streams : events arrives from the past. You can check [this test case](https://github.com/riemann/riemann/pull/789/files#diff-f4d4815a1ac7dab86942b0dda98a1d7cR1764) for example.

What do you think about these streams ? Do you think they have their place in Riemann ?

### DIFF
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -2089,3 +2089,72 @@
                             (dissoc ::clock-skew-timestamp)
                             (assoc :metric delta))]
                 (call-rescue event children)))))))
+
+(defn cond-dt
+  "A stream which detects if a condition `(f event)` is true during `dt` seconds.
+  Takes a function of events `f` and a time period `dt` in seconds.
+&
+    If the condition is valid for all events received during at least the period `dt`, valid events received after the `dt` period will be passed on until an invalid event arrives.
+  Skips events that are too old or that do not have a time stamp."
+  [f dt & children]
+  (let [last-changed-state (atom {:ok false
+                                  :time nil})]
+    (fn [event]
+      (let [{ok :ok changed-state-time :time} @last-changed-state
+            event-time (:time event)
+            valid-event (f event)]
+        (when event-time ;; filter events with no time
+          (swap! last-changed-state (fn [state]
+                        (cond
+                          (and valid-event (and (not ok)
+                                                (or (not changed-state-time)
+                                                    (> event-time changed-state-time))))
+                          ;; event is validating the condition
+                          ;; last event is not ok, has no time or is too old
+                          ;; => last-changed-state is now ok with a new time
+                          {:ok true :time event-time}
+                          (and (not valid-event) (and ok
+                                                      (or (not changed-state-time)
+                                                          (> event-time changed-state-time))))
+                          ;; event is not validating the condition
+                          ;; last event is ok, has no time or is too old
+                          ;; => last-changed-state is now ko with a new time
+                          {:ok false :time event-time}
+                          ;; default value, return the state
+                          :default state)))
+          (when (and valid-event
+                      ;; we already had an ok event
+                     ok
+                     ;; check is current time > first ok event + dt
+                     (> event-time (+ changed-state-time dt)))
+            (call-rescue event children)))))))
+
+(defn above
+  "Takes a number `threshold` and a time period in seconds `dt`.
+  If the condition `(> (:metric event) threshold)` is valid for all events received during at least the period `dt`, valid events received after the `dt` period will be passed on until an invalid event arrives.
+  `:metric` should not be nil (it will produce exceptions)."
+  [threshold dt & children]
+  (apply cond-dt (fn [event] (> (:metric event) threshold)) dt children))
+
+(defn below
+  "Takes a number `threshold` and a time period in seconds `dt`.
+  If the condition `(< (:metric event) threshold)` is valid for all events received during at least the period `dt`, valid events received after the `dt` period will be passed on until an invalid event arrives.
+  `:metric` should not be nil (it will produce exceptions)."
+  [threshold dt & children]
+  (apply cond-dt (fn [event] (< (:metric event) threshold)) dt children))
+
+(defn between
+  "Takes two numbers, `low` and `high`, and a time period in seconds, `dt`.
+  If the condition `(and (> (:metric event) low) (< (:metric event) high))` is valid for all events received during at least the period `dt`, valid events received after the `dt` period will be passed on until an invalid event arrives.
+  `:metric` should not be nil (it will produce exceptions)."
+  [low high dt & children]
+  (apply cond-dt (fn [event] (and (> (:metric event) low)
+                                  (< (:metric event) high))) dt children))
+
+(defn outside
+  "Takes two numbers, `low` and `high`, and a time period in seconds, `dt`.
+  If the condition `(or (< (:metric event) low) (> (:metric event) high))` is valid for all events received during at least the period `dt`, valid events received after the `dt` period will be passed on until an invalid event arrives.
+  `:metric` should not be nil (it will produce exceptions)."
+  [low high dt & children]
+  (apply cond-dt (fn [event] (or (< (:metric event) low)
+                                 (> (:metric event) high))) dt children))
--- a/test/riemann/streams_test.clj
+++ b/test/riemann/streams_test.clj
@@ -1691,3 +1691,248 @@
                                  {:time 9 :host :baz :metric 5}
                                  {:time 1 :host :foo :metric -4}
                                  {:time 100 :host :foo :metric 89}]))
+
+(deftest above-test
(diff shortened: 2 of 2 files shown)


src/riemann/streams.clj
```diff
@@ -2089,3 +2089,67 @@
...
+                          :default state)))
+          (when (and valid-event
+                      ;; we already had an ok event
+                     ok
+                     ;; check is current time > first ok event + dt
+                     (> event-time (+ changed-state-time dt)))
+            (call-rescue event children)))))))
+
+(defn above
+  "Takes a number `threshold` and a time periode in seconds `dt`. Detects if the `:metric` value is > to thresold during dt seconds using cond-dt.
```
boernd: Typos (periode and thresold)

mcorbin: fixed, thanks

src/riemann/streams.clj
```diff
@@ -2089,3 +2089,66 @@
                             (dissoc ::clock-skew-timestamp)
                             (assoc :metric delta))]
                 (call-rescue event children)))))))
+
+(defn cond-dt
+  "Takes a function of events `f` and a time period `dt` in seconds. Detects if `(f event)` is true during dt seconds.
```
Mogztter: I think the documentation is a bit misleading.
What about:

> A stream which detects if a condition `(f event)` is true during `dt` seconds.
Takes a function of events `f` and a time period `dt` in seconds.

> If the condition is valid for all events received during at least the period `dt`, valid events received after the `dt` period will be passed on until an invalid event arrives.
Skips too old events or events without time.

Disclaimer: I'm not a native English speaker :wink:

mcorbin: I like your description, i updated the PR ;) Thanks !

src/riemann/streams.clj
```diff
@@ -2089,3 +2089,68 @@
                             (dissoc ::clock-skew-timestamp)
                             (assoc :metric delta))]
                 (call-rescue event children)))))))
+
+(defn cond-dt
+  "A stream which detects if a condition `(f event)` is true during `dt` seconds.
+  Takes a function of events `f` and a time period `dt` in seconds.
+&
+    If the condition is valid for all events received during at least the period `dt`, valid events received after the `dt` period will be passed on until an invalid event arrives.
+    Skips too old events or events without time."
```
jamtur01: `Skips events that are too old or that do not have a time stamp.`

src/riemann/streams.clj
```diff
@@ -2089,3 +2089,68 @@
...
+          (when (and valid-event
+                      ;; we already had an ok event
+                     ok
+                     ;; check is current time > first ok event + dt
+                     (> event-time (+ changed-state-time dt)))
+            (call-rescue event children)))))))
+
+(defn above
+  "Takes a number `threshold` and a time period in seconds `dt`. Detects if the `:metric` value is > to threshold during dt seconds using cond-dt.
+  :metric should not be nil."
```
jamtur01: What happens if the `:metric` is nil? Perhaps explain that?

src/riemann/streams.clj
```diff
@@ -2089,3 +2089,68 @@
...
+                          :default state)))
+          (when (and valid-event
+                      ;; we already had an ok event
+                     ok
+                     ;; check is current time > first ok event + dt
+                     (> event-time (+ changed-state-time dt)))
+            (call-rescue event children)))))))
+
+(defn above
+  "Takes a number `threshold` and a time period in seconds `dt`. Detects if the `:metric` value is > to threshold during dt seconds using cond-dt.
```
jamtur01: `the `:metric` value is > to threshold during dt seconds using cond-dt` - I don't understand what you're trying to say here?

Also inconsistent use of back tick quoting.

src/riemann/streams.clj
```diff
@@ -2089,3 +2089,68 @@
...
+  :metric should not be nil."
+  [threshold dt & children]
+  (apply cond-dt (fn [event] (> (:metric event) threshold)) dt children))
+
+(defn below
+  "Takes a number `threshold` and a time period in seconds `dt`. Detects if the `:metric` value is < to threshold during dt seconds using cond-dt.
```
jamtur01: As above.

src/riemann/streams.clj
```diff
@@ -2089,3 +2089,68 @@
...
+  :metric should not be nil."
+  [threshold dt & children]
+  (apply cond-dt (fn [event] (< (:metric event) threshold)) dt children))
+
+(defn between
+  "Takes two numbers `low` and `high` and a time period in seconds `dt`. Detects if the `:metric` value is > to low AND < to high during dt seconds using cond-dt.
```
jamtur01: "Takes two numbers, `low` and `high`, and a time period in seconds, `dt`."

jamtur01: Also inconsistent quoting.

jamtur01: "Detects if the `:metric` value is > to low AND < to high during dt seconds using cond-dt." - again unclear....

src/riemann/streams.clj
```diff
@@ -2089,3 +2089,68 @@
...
+  :metric should not be nil."
+  [low high dt & children]
+  (apply cond-dt (fn [event] (and (> (:metric event) low)
+                                  (< (:metric event) high))) dt children))
+
+(defn outside
```
jamtur01: As above.

src/riemann/streams.clj
mcorbin: Fixed, thanks !

src/riemann/streams.clj
mcorbin: I rewrote the description. Sorry for my bad english :) Thanks !

src/riemann/streams.clj
mcorbin: I explained it, thanks !

src/riemann/streams.clj
mcorbin: I modified the description, thanks !

src/riemann/streams.clj
mcorbin: Fixed, thanks !

mcorbin: Fixed, thanks !

mcorbin: I modified the description, thanks !

src/riemann/streams.clj
mcorbin: Fixed, thanks !

mcorbin:
@aphyr @pyr What do you think about it ? I think the use case "is (f event) true during dt seconds" is very common. I tried to do a clean implementation of the functionality, but maybe we can do better ?

aphyr:
So... I don't doubt that this is useful, but as far as Riemann's core goes, I try to keep functions limited to those that will have broad applicability with obvious, composable semantics. These seem like... specific sorts of functions that have lots of variants depending on use case, and I think the best thing to do there is to let people implement them themselves via (fixed-time-window (smap (fn [events] (every? pred event))) or similar.

mcorbin:
The problem i have with `(fixed-time-window (smap (fn [events] (every? pred event)))` is this:

|----TW1----|----TW2---|              
|ooooXXXX|XXXXoooo|
|-----------------------------> t

I will miss the problem because it happens in 2 time windows. 

But i can also release it as a plugin ;)

aphyr:
Or moving-time-window, or sreduce directly.

mcorbin:
Ok, i close it (and will move the code in a plugin). Thanks you ;)

PR #790, opened 2017-03-21
uses humane-test-output to pretty print test outputs

### LINKED ISSUE #787: Pretty print test output
Tests outputs are difficult to read.

It should be feasible to modify it using https://github.com/clojure/clojure/blob/master/src/clj/clojure/test.clj#L211.

mcorbin:
I just discovered [humane-test-output](https://github.com/pjstadig/humane-test-output).
Before using it :

```
FAIL in (foo-test) (riemann.config:25)
expected: (= (:foo result) [{:host "localhost", :service "foo1", :metric 10}])
  actual: (not (= [{:host "localhost", :service "foo", :metric 10}] [{:host "localhost", :service "foo1", :metric 10}]))
```

after :

```
FAIL in (foo-test) (riemann.config:25)
expected: [{:host "localhost", :service "foo", :metric 10}]
  actual: [{:host "localhost", :service "foo1", :metric 10}]
    diff: - [{:service "foo"}]
          + [{:service "foo1"}]
```

this PR fix #787.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -43,7 +43,8 @@
     [clj-campfire "2.2.0"]
     [clj-nsca "0.0.3"]
     [amazonica "0.3.28" :exclusions [joda-time]]
-    [spootnik/kinsky "0.1.16"]]
+    [spootnik/kinsky "0.1.16"]
+    [pjstadig/humane-test-output "0.8.1"]]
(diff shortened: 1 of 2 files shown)


jamtur01:
LGTM

jamtur01:
Thanks!

PR #797, opened 2017-04-02
replace refs in fixed-time-window-fn by an atom

mcorbin:
fixed-time-window-fn used 2 refs (start-time and buffer) to track
states. 
This PR replace them by an atom, which should improve performance.

### DIFF
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -324,7 +324,6 @@
   *not* overlap; each event appears at most once in the output stream. Once an
   event is emitted, all events *older or equal* to that emitted event are
   silently dropped.
-
   Events without times accrue in the current window."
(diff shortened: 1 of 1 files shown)


mcorbin:
Some numbers, using riemann-bench on my laptop with this configuration:

```clojure

(streams
 (default :ttl 300 (index))
 (fixed-time-window 2)
 (fixed-time-window 3)
 (fixed-time-window 4)
 (fixed-time-window 5)
 (fixed-time-window 6)
 (fixed-time-window 7)
 (fixed-time-window 8)
 (fixed-time-window 9))
```


bulk latency
single latency
single throughtput
bulk throughtput

jamtur01:
And this too! @mcorbin for MVP.

PR #804, opened 2017-04-07
Refactoring Elasticsearch output

### LINKED ISSUE #791: [feature request] Support batch functionality in elasticsearch output
As far as I can see in the code the elasticsearch output just accepts single events and does not support the bulk api of elasticsearch.

mcorbin:
Fixes #791

The current Elasticsearch output does not change.

I added a new steam `elasticsearch-bulk` to support Elasticsearch bulk API.

Stream options are:

```
:es-endpoint     Elasticsearch, default is \"http://127.0.0.1:9200\".
:username        Username to authenticate with.
:password        Password to authenticate with.
:formatter       Fn taking an event and returning it with the ES Bulk API format
:http-options    Http options (like proxy). See https://github.com/dakrone/clj-http for option list
```
The stream must receive events with this format:

```clojure
  {:es-action "index"
   :es-metadata {:_index "test"
                 :_type "type1"
                 :_id "1"}
   :es-source {:field1 "value1"}}
```
(note : It's exactly the format expected by ES bulk API, except for the `es-action` key : https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html)

Or you can use the :formatter parameter to pass a function that will convert events from the Riemann format to the Elasticsearch format.

I have added a generic formatter (`default-bulk-formatter`), which should be sufficient for most use cases:

```
(defn default-bulk-formatter
  "Returns a function which accepts an event and format it for the Elasticsearch bulk API.

  Options :
  :es-index        Elasticsearch index name (without suffix).
  :type            Type to send to index.
  :es-action       Elasticsearch action, for example \"index\".
  :index-suffix    Index-suffix, for example \"-yyyy.MM.dd\".

  Each event received by the function can also have these keys (which override default options), and a optional `es-id` key."
```

For example:

```clojure
    (let [formatter (default-bulk-formatter {:es-index "riemann"
                                             :type "foo"
                                             :es-action "index"
                                             :index-suffix "-yyyy.MM.dd"})
          elastic (elasticsearch-bulk {:username "elastic"
                                       :password "changeme"
                                       :formatter formatter})]
      (elastic {:host "fooformatter"
                :metric 10
                :service "bar"
                :tags ["t1"]
                :ttl 30
                :time 1491567382
                :foo "bar"}))
```
You will have in Elasticsearch:

```json
{
  "_index": "riemann-2017.04.07",
  "_type": "foo",
  "_id": "AVtIjyj5EApgwWFtAOGj",
  "_score": 1,
  "_source": {
    "host": "fooformatter",
    "metric": 10,
    "service": "bar",
    "tags": [
      "t1"
    ],
    "ttl": 30,
    "foo": "bar",
    "@timestamp": "2017-04-07T12:16:22.000Z"
  },
  "fields": {
    "@timestamp": [
      1491567382000
    ]
  }
}
```

With this stream and this formatter, you can specify per event id, index, type, action (create, update...). It's also easy to write your own formatter if you need it.

I tested this PR using Elsaticsearch 5.3 and 2.4.

I want @boernd and @faxm0dem reviews on this PR before merging (and i also want to make more tests).

### DIFF
--- a/src/riemann/elasticsearch.clj
+++ b/src/riemann/elasticsearch.clj
@@ -22,16 +22,17 @@
 
 (defn- post
   "POST to Elasticsearch."
-  [credentials esindex formatted-event]
-  (let [base-http-options {:body (json/generate-string formatted-event)
-                           :content-type :json
-                           :conn-timeout 5000
-                           :socket-timeout 5000
-                           :throw-entire-message? true}
+  [credentials es-endpoint body http-options]
+  (let [base-http-options (merge {:body body
+                                  :content-type :json
+                                  :conn-timeout 5000
+                                  :socket-timeout 5000
+                                  :throw-entire-message? true}
+                                 http-options)
         http-options (if credentials
                        (assoc base-http-options :basic-auth credentials)
                        base-http-options)]
-    (http/post esindex http-options)))
+    (http/post es-endpoint http-options)))
 
 (defn elasticsearch
   "Returns a function which accepts an event and sends it to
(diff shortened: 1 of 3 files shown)


faxm0dem:
I'll test this on monday, thanks a bunch!

src/riemann/elasticsearch.clj
```diff
@@ -59,20 +60,120 @@
...
+  (reduce gen-request-bulk-body-reduce "" events))
+
+(defn default-bulk-formatter
+  "Returns a function which accepts an event and format it for the Elasticsearch bulk API.
+
+  Options :
+  :es-index        Elasticsearch index name (without suffix).
+  :type            Type to send to index.
+  :es-action       Elasticsearch action, for example \"index\".
+  :index-suffix    Index-suffix, for example \"-yyyy.MM.dd\".
```
boernd: typo -> Index suffix

src/riemann/elasticsearch.clj
```diff
@@ -59,20 +60,120 @@
...
+
+  Each event received by the function can also have these keys (which override default options), and a optional `es-id` key."
```
boernd: typo -> ... and an optional

src/riemann/elasticsearch.clj
```diff
@@ -59,20 +60,120 @@
...
+                       (assoc m :_id es-id)
+                       m))]
+      {:es-action es-action
+       :es-metadata metadata
+       :es-source source})))
+
+(defn elasticsearch-bulk
+  "Returns a function which accepts an event (or a list of events) and sends it to
+  Elasticsearch using the Bulk API. Custom event formatter can be provided using the `:formatter` key.
+  A formatter is a function wihci accepts an event.
```
boernd: typo -> which

src/riemann/elasticsearch.clj
```diff
@@ -59,20 +60,120 @@
...
+
+  Events should have this format :
+
+  {:es-action \"index\"
+   :es-metadata {:_index \"test\"
+                 :_type \"type1\"
+                 :_id \"1\"}
+   :es-source {:field1 \"value1\"}}
+
+  `:es-action` is the action (create, update, index delete), `:es-metadata` the document metadata, and `es-source` the document source.
```
boernd: - typo ... update, index, delete...
- I don't think the update action would work at the moment because of the doc syntax:
````
{ "update" : {"_id" : "1", "_type" : "type1", "_index" : "test"} }
{ "doc" : {"field2" : "value2"} }
````
Maybe just document that update is not supported? I do not see much of a use case with Riemann here

**Comment by boernd** [comment]
Thanks @mcorbin ! Will also test this week...

File: src/riemann/elasticsearch.clj [hunk 5]
```diff
@@ -59,20 +60,120 @@
...
+   (when (:es-source elem)
+     (str (json/generate-string (:es-source elem)) "\n"))))
+
+(defn gen-request-bulk-body
+  "Takes a list of events, generates the body request for Elasticsearch"
+  [events]
+  (reduce gen-request-bulk-body-reduce "" events))
+
+(defn default-bulk-formatter
+  "Returns a function which accepts an event and format it for the Elasticsearch bulk API.
```
mcorbin: Fixed, thanks ;)

src/riemann/elasticsearch.clj
mcorbin: Fixed, thanks ;)

src/riemann/elasticsearch.clj
mcorbin: Fixed, thanks ;)

src/riemann/elasticsearch.clj
mcorbin: Fixed, thanks ;)

src/riemann/elasticsearch.clj
mcorbin: I fixed the typo, thanks.
Update will works is the doc syntax is used (i tested it). You could write a custom formatter generating the doc syntax if the operation is update.

faxm0dem:
I just tested this against Searchguard-enabled ES 2.4.1 in HTTPS.
I love the semantics of your stream, and it works perfectly.
I'll integrate it with [samplerr](/ccin2p3/samplerr) ASAP to see how it performs although I probably won't have time before May :-/

mcorbin:
Glad you like the stream ;)

- I could add in logfile the number of processed messages but actually no stream does it. If i add it it will be disabled by default.
- The stream doesn't retry but i can add it (don't really know how actually, i don't want to block the thread too long with multiple retries).

I will also try to take a look at samplerr next week.

faxm0dem:
don't bother logging if it's non-standard. However, internal stats would help: created/updated/failed. Then again I guess if something fails, it would make it into the exception stream, wouldn't it?

boernd:
Running with ES 5.1.2, no issues so far.

Shouldn't the `elasticsearch-bulk` and `default-bulk-formatter` functions be referred into the `riemann.config` ns like the old `elasticsearch` function?

mcorbin:
I added elasticsearch-bulk and default-bulk-formatter on riemann.config, thanks ;)
@faxm0dem Yes, if something fails, you will have an exception. I can add (optional) internal stats, i will think about it.

boernd:
@Retry: I had a look at the logstash elasticsearch output, they retry indefinitely (https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/common.rb). I'm not aware that any Riemann output does retry atm? 

```
# Retry logic: If it is success, we move on. If it is a failure, we have 3 paths:
# - For 409, we log and drop. there is nothing we can do
# - For a mapping error, we send to dead letter queue for a human to intervene at a later point.
# - For everything else there's mastercard. Yep, and we retry indefinitely. This should fix #572 and other transient network issues
```

Anything else open to get this merged?

mcorbin:
@faxm0dem : i prefer not to add internal stats actually because i don't know the best way to implement it.
@boernd : Maybe a retry stream "(retry number children)", catching exceptions, will be useful ?
I think we can merge.

faxm0dem:
np we can add them later!

jamtur01:
Thanks!

PR #808, opened 2017-04-12
Extend pushover functionality

boernd:
Support
- custom event formatter
- pushover parameters

I did not want to break compatibility so I left the `token` and `user` parameters untouched.


jamtur01:
LGTM. Thanks!

PR #821, opened 2017-06-05
Mock the Riemann index in test mode

mcorbin:
This PR adds the ability to mock the Rieman index in test mode. I saw this feature requested several times on irc/mailing list/issues.

## How

I defined a `^:dynamic *index*` variable in riemann.test. This variable will contains an index in test mode (set by `with-test-env`). The index will be cleared between each `deftest` (you can see it in the `deftest` macro in riemann.test).

I modified `riemann.config/index` (replaced it by a macro). It test mode, it will returns a stream (`riemann.test/index-stream`) which will record events in the test index. Events without time are not indexed (but it will not cause exceptions).

I also added several utility functions in riemann.test :

- lookup-index: "Return an indexed event from the test index"
- clear-index: "Clear the test index."
- query-index: "Return a vec of events, sorted by `:host` and `:service`, from the test index matching the query."
- expire-index: "Return a vec of expired events, sorted by `:host` and `:service`, from the test index, removing each."

You can use this functions in your tests (i will show some examples later). `query-index` and `expire-index` returns sorted vector, so their results can be easily compared with expected results in tests.

I also added `(reset! riemann.config/core {:index *index*})` in `with-test-env`. Streams querying the index (like [the maintenance mode stream](http://riemann.io/howto.html#query-the-index-from-within-a-stream)) should now be testable (it was not the case because there is no core in test mode).

## Commented examples

Test indexed events :

```
(streams
 ;; index all events with hosts = foo
 (where (host "foo")
  (index)))

(tests
 (deftest foo
   (inject! [{:host "foo" :service "a" :metric 0 :time 0 :ttl 30}
             {:host "foo" :service "b" :metric 0 :time 0 :ttl 30}
             {:host "bar" :service "c" :metric 0 :time 0 :ttl 30}
             {:host "foo" :service "c" :metric 0 :time 0 :ttl 30}])
   (is (= (query-index "true")
          ;; use query-index to query the index
          [{:host "foo" :service "a" :metric 0 :time 0 :ttl 30}
           {:host "foo" :service "b" :metric 0 :time 0 :ttl 30}
           {:host "foo" :service "c" :metric 0 :time 0 :ttl 30}]))

   (is (= (lookup-index "foo" "a")
          ;; use lookup-index to retrieve an event by host/service
          {:host "foo" :service "a" :metric 0 :time 0 :ttl 30}))
   ;; advance time to expire events
   (riemann.time.controlled/advance! 31)
   ;; test for expired events
   (is (= (expire-index)
          ;; events are expired
          [{:host "foo" :service "a" :time 31 :state "expired"}
           {:host "foo" :service "b" :time 31 :state "expired"}
           {:host "foo" :service "c" :time 31 :state "expired"}]))
   ;; the index is empty, all events are expired
   (is (= (query-index "true") []))

   ;; inject some new events
   (inject! [{:host "foo" :service "a" :metric 0 :time 30 :ttl 30}
             {:host "foo" :service "b" :metric 0 :time 30 :ttl 30}
             {:host "foo" :service "c" :metric 0 :time 30 :ttl 30}])
   (is (= (query-index "host = \"foo\"")
          ;; query the index for host = foo
          [{:host "foo" :service "a" :metric 0 :time 30 :ttl 30}
           {:host "foo" :service "b" :metric 0 :time 30 :ttl 30}
           {:host "foo" :service "c" :metric 0 :time 30 :ttl 30}]))
   ;; advance time to expire events
   (riemann.time.controlled/advance! 61)

   (is (= (expire-index {:keep-keys [:host :service :metric]})
          ;; use the :keey-keys option like in periodically-expire
          [{:host "foo" :service "a" :metric 0 :time 61 :state "expired"}
           {:host "foo" :service "b" :metric 0 :time 61 :state "expired"}
           {:host "foo" :service "c" :metric 0 :time 61 :state "expired"}]))))
```

Test [the maintenance mode stream](http://riemann.io/howto.html#query-the-index-from-within-a-stream).

```clojure
(defn maintenance-mode?
  "Is Riemann currently in maintenance mode?"
  []
  ; Take an expression representing a query for maintenance mode
  (->> '(and (= :host nil)
             (= :service "maintenance-mode"))
       ; Search the current Riemann core's index for any matching events
       (riemann.index/search (:index @core))
       ; Take the first match
       first
       ; Find its state
       :state
       ; Is it the string "active"?
       (= "active")))

(streams
 (where (host "maintenance")
   ;; index a maintenance event
   (with {:host nil
          :service "maintenance-mode"}
     (index)))
 (where (and (not (host "maintenance"))
             (not (maintenance-mode?)))
   ;; index all events excepts maintenance events
   (tap :maintenance-mode)))

(tests
 (deftest maintenance-test
   (let [result (inject! [{:host "foo" :service "a" :time 0}
                          {:host "bar" :service "a" :time 1}
                          ;; enable maintenance
                          {:host "maintenance" :state "active" :time 2}
                          ;; these events will not be present in the tap
                          {:host "bar" :service "b" :time 3}
                          {:host "bar" :service "c" :time 4}
                          ;; disable maintenance
                          {:host "maintenance" :state "not active" :time 5}
                          {:host "bar" :service "d" :time 6}])]
     (is (= (:maintenance-mode result)
            [{:host "foo" :service "a" :time 0}
             {:host "bar" :service "a" :time 1}
             {:host "bar" :service "d" :time 6}])))))
```

I think it's important that someone give this PR a second look before merging.

### DIFF
--- a/src/riemann/config.clj
+++ b/src/riemann/config.clj
@@ -206,7 +206,7 @@
     (swap! next-core assoc :streams
            (reduce conj (:streams @next-core) things))))
 
-(defn index
+(defn set-index
   "Set the index used by this core. Returns the index."
   [& opts]
   (locking core
@@ -223,6 +223,13 @@
                    index')]
       (swap! next-core assoc :index index')
       index')))
+
+(defmacro index
+  "Set the index used by this core. Returns the index."
+  [& opts]
+  (if test/*testing*
+    `(test/index-stream)
+    `(apply set-index ~opts)))
 
 (defn update-index
   "Updates the given index with all events received. Also publishes to the
--- a/src/riemann/test.clj
+++ b/src/riemann/test.clj
@@ -7,6 +7,7 @@
   received."
   (:require [riemann.time.controlled :as time.controlled]
             [riemann.time :as time]
+            [riemann.index :as index]
             [riemann.streams :as streams]
             [clojure.test :as test]))
 
@@ -30,6 +31,10 @@
   time."
   nil)
 
+(def ^:dynamic *index*
+  "Test index"
+  false)
+
 (defn tap-stream
   "Called by `tap` to construct a stream which records events in *results*
   before forwarding to child."
@@ -42,6 +47,45 @@
 
     ; Forward downstream
     (child event)))
+
+(defn index-stream
+  "Called by `riemann.config/index` to construct a stream which records events in *index*.
+  Do not index events with no time."
+  []
+  (fn [event]
+    (when (:time event)
+      (index/insert *index* event))))
+
+(defn lookup-index
+  "Return an indexed event from the test index"
+  [host service]
+  (index/lookup *index* host service))
+
+(defn clear-index
+  "Clear the test index."
+  []
+  (index/clear *index*))
+
+(defn query-index
+  "Return a vec of events, sorted by `:host` and `:service`, from the test index matching the query."
+  [query]
+  (->> (index/search *index* (riemann.query/ast query))
+      (sort-by #(vec (map % [:host :service])))
+      (vec)))
+
+(defn expire-index
+  "Return a vec of expired events, sorted by `:host` and `:service`, from the test index, removing each."
+  ([] (expire-index {}))
+  ([opts]
+   (let [keep-keys (get opts :keep-keys [:host :service])
+         expired-events (index/expire *index*)]
+     (->> (map (fn [event]
+                 (-> (select-keys event keep-keys)
+                     (merge {:state "expired"
+                             :time (time/unix-time)})))
+               expired-events)
+          (sort-by #(vec (map % [:host :service])))
+          (vec)))))
 
 (defmacro tap
   "A stream which records inbound events in the *results* map. Takes a globally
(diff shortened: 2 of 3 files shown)


jamtur01:
This touches more than I am comfortable +1'ing - @pyr @aphyr's call.

aphyr:
I don't think this is a good idea. Yes, index support in tests is something that people ask for, but the proposed changes introduce significant complexity while not solving the underlying problem. Notably:

- It introduces a macro for index, which breaks (apply).
- It assumes there is only one implementation for indices, but Index is a polymorphic protocol. Riemann has at least two implementations, and users run more.
- The test index does not behave like any real index. Tests should be faithful to the thing they're representing.
- It does not support the reaper, which is the most common pain point for users of riemann.test.
- It adds index-specific features and more global state to the test namespace.
- It introduces an incomplete core. I think there should be either no core, or a core that resembles the full one.
- Manual expiry introduces another way for user tests to diverge from true index behavior.

I strongly feel that adding support for testing service behavior, like indices and the reaper, relies on a full testing implementation of a core, with comprehensive control of side effects throughout *all* Riemann services. That's a nontrivial project, but I think it's better to be explicit about the lack of support than to provide broken, incomplete support for indices.

PR #825, opened 2017-06-06
Add sflatten stream

mcorbin:
When i started using Riemann, i quickly needed a stream to convert seq of events to individual events. This is what this stream does, and i think it can be useful to others. Example :

```clojure
(streams
  (fixed-time-window 10
    (sflatten
      #(info %))))
```

### DIFF
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -182,6 +182,13 @@
     (doseq [e (f event)]
       (call-rescue e children))))
 
+(defn sflatten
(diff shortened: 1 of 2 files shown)


faxm0dem:
very useful indeed :+1:

jamtur01:
I am okay - @pyr any thoughts?

jamtur01:
Also @aphyr.

aphyr:
Works for me. You can also express this as (smapcat identity). :)

jamtur01:
Awesome.

PR #829, opened 2017-06-12
Remove all *warn-on-reflection* warning in the influxdb stream

mcorbin:
Remove all *warn-on-reflection* warning in the influxdb stream.


jamtur01:
Thanks!

PR #831, opened 2017-06-21
Netuitive Payload Size Reduction

TheConnMan:
Netuitive accepts REST payloads containing arrays of elements, but we prefer having a smaller number of larger payloads. The current implementation sends element payloads with a single metric and a single sample which are very inefficient.

This PR batches element payloads so a typical REST payload will have a few unique elements, a few metrics each, and a lot of samples. The PR also filters out events with a sample value of `nil` because those are ignored anyways. The resulting data in Netuitive will be identical, this just more efficiently transfers the data.

### DIFF
--- a/src/riemann/netuitive.clj
+++ b/src/riemann/netuitive.clj
@@ -77,5 +77,6 @@
     (fn [event]
       (let [events (if (sequential? event) event [event])
             post-data (mapv #(generate-event % opts) events)
-            json-data (generate-string post-data)]
+            reduced-data (map (apply partial reduce combine-elements []) (partition-by :id (sort-by :id post-data)))
+            json-data (generate-string reduced-data)]
         (post-datapoint (:api-key opts) (:url opts gateway-url) json-data)))))
--- a/src/riemann/netuitive.clj
+++ b/src/riemann/netuitive.clj
@@ -34,11 +34,11 @@
    "Combine two elements"
    [element1 element2]
    {:id (:id element1)
-     :name (:name element1)
-     :type (:type element1)
-     :metrics (map (partial apply merge) (partition-by :id (sort-by :id (concat (:metrics element1) (:metrics element2)))))
-     :samples (concat (:samples element1) (:samples element2))
-     :tags (map (partial apply merge) (partition-by :name (sort-by :name (concat (:tags element1) (:tags element2)))))})
+    :name (:name element1)
+    :type (:type element1)
+    :metrics (map (partial apply merge) (partition-by :id (sort-by :id (concat (:metrics element1) (:metrics element2)))))
+    :samples (concat (:samples element1) (:samples element2))
+    :tags (map (partial apply merge) (partition-by :name (sort-by :name (concat (:tags element1) (:tags element2)))))})
 
 (defn generate-event
    "Structure for ingest to Netuitive as JSON"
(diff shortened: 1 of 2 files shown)


src/riemann/netuitive.clj
```diff
@@ -30,6 +30,16 @@
    [tag]
    {:name tag :value :true})
 
+(defn combine-elements
+   "Combine two elements"
+   [element1 element2]
+   {:id (:id element1)
+     :name (:name element1)
```
mcorbin: You should remove one space before ":"

TheConnMan: On 37-41, correct?

mcorbin: Yes, it should be align with :id ;)

TheConnMan: Hey, it's my first Clojure PR, anything's possible.

mcorbin:
Hi,

Could you write a couple of tests showing/testing the complete output of `combine-elements` ? I want to be sure to completely understand what is the expected result ;)

TheConnMan:
Yes I can, just pushed some more explicit tests showing that:
- Metric arrays are merged by distinct ID
- Tag arrays are merged by key
- Sample arrays are concatenated

mcorbin:
I don't understand why you use :sort, is the order of `:metrics` and `:tags` important ?

If you just want to have unique `:metrics` and `:tags`, you could use set, example:

```clojure

(defn combine-elements
   "Combine two elements"
  [element1 element2]
  (assoc element1 :metrics (clojure.set/union (:metrics element1) (:metrics element2))
                  :samples (concat (:samples element1) (:samples element2))
                  :tags    (clojure.set/union (:tags element1) (:tags element2))))

(defn generate-event
   "Structure for ingest to Netuitive as JSON"
   [event opts]
   (let [type (:type opts "Riemann")
         metric-id (netuitive-metric-name event)]
       {:id (str type ":" (:host event))
        :name (:host event)
        :type type
        :metrics (set [{:id metric-id}])
        :samples [{:metricId metric-id
                   :timestamp (parsetime (:time event))
                   :val (:metric event)}]
        :tags (set (map generate-tag (:tags event)))}))
```

The `netuitive fn` can also be modified :

```clojure
(let [opts (merge {:api-key "netuitive-api-key"} opts)]
    (fn [event]
      (let [json-data (->> (if (sequential? event) event [event])
                           (filter :metric)
                           (map #(generate-event % opts))
                           (partition-by :id)
                           (map #(reduce combine-elements %))
                           (generate-string))]
        (post-datapoint (:api-key opts) (:url opts gateway-url) json-data))))
```

We can probably use transducers to have better performances.

TheConnMan:
You're entirely right about not needing the sort, they do not need to be sorted. I'll make the modifications you've recommended locally, do a bit of testing, and push them. Thank you for the detailed feedback.

TheConnMan:
Worked like a charm. I verified that the resulting Netuitive payloads are still smaller and accurate. Thanks for the Clojure tricks!

mcorbin:
In my example, i don't use transducers but threading macro, but it's a good solution too ;)

LGTM

TheConnMan:
Ah, my mistake. Thanks for all the quick feedback!

jamtur01:
Thanks!

TheConnMan:
@jamtur01 How often do you release this project? I'm looking forward to deploying the updated build artifacts.

jamtur01:
I'll try to do a release soon. Time constraints I am afraid.

jamtur01:
It's deployed.

TheConnMan:
Thank you, much appreciated!

PR #837, opened 2017-07-22
Add Pagerduty v2 API support

mcorbin:
By default, the pagerduty stream uses the v1 API (so no breaking change), but the v2 API can be used by setting the `:version` option to `:v2`.

You can find more informations about the v2 API [here](https://v2.developer.pagerduty.com/docs/events-api-v2).

When you use the v2 API, each event can also have a `:dedup-key` field. It will be used by Pagerduty for alert de-duplication (cf v2 api documentation).

cc @openglx ;)


jamtur01:
LGTM!

PR #840, opened 2017-07-27
Upgrade influxdb-java dependency

mcorbin:
influxdb-java changelog [here](https://github.com/influxdata/influxdb-java/blob/master/CHANGELOG.md).

Two nice improvements:

```
Significant (~35%) performance improvements for write speed with less memory footprint. PR #330
Drop guava runtime dependency which reduces jar size from 1MB -> 49KB PR #322
```


mcorbin:
This version of influxdb-java is not compatible with Java 7. Is it a problem ? I think we should stop support Java 7.

jamtur01:
Blocked by #846

PR #842, opened 2017-08-01
Replace Throwable by Exception

### LINKED ISSUE #833: riemann.streams/call-rescue is catching java.lang.Error
call-rescue catches and handles java errors like OutOfMemoryError. This will put JVM in unpredictable state.

mcorbin:
Fix #833.

Why catch Throwable ?


jamtur01:
Thanks!

Issue #848, opened 2017-08-24
[regression] InfluxDB lib should be fed consistent data type


XANi:
After upgrading 0.2.11 -> 0.2.14 we've started getting errors:

    java.lang.RuntimeException: {"error":"partial write: field type conflict: input field \"value\" on measurement \"riemann\" is type integer, already exists as type float dropped=3"}

I've noticed that when some of counters are zero, they are sent to influxdb as integer (influxdb lib generates value with`i` at the en of it, then when they are increased, they are sent as float:

    riemann,host=hal1,plugin_instance=longterm,type=gauge,type_instance=accepted value=0.199541055572184 1503596351000000000
    riemann,host=hal1,plugin_instance=longterm,type=gauge,type_instance=completed value=0.199541055572184 1503596351000000000
    riemann,host=hal1,plugin_instance=longterm,type=gauge,type_instance=rejected value=0i 1503596351000000000

those are just internal riemann counters wrapped in `(fixed-time-window 10 (smap folds/mean) graph-default)` (graph-default is just my helper that calls influxdb with right options).

We've also had different error (altho that one doesn't show up as reliably):

    unable to parse 'protocols,aggregate=max,host=efikom116.non.3dart.com,plugin_instance=Tcp,type=protocol_counter,type_instance=CurrEstab value=� 1503596468000000000': invalid boolean\n

which seems to be related to influxdata/influxdb-java#39

but on wire (tcpdump) it looks like that

     protocols,aggregate=mean,host=efikom116.non.3dart.com,plugin_instance=TcpExt,type=protocol_counter,type_instance=TCPTimeouts value=... 1503597973000000000

value morphed to `...` somehow...

Curiously enough, that one happened only when I restarted source of events (collectd) and it sent all plugin output at once and *stopped* when I (re-)added batching (via `(batch 10000 10 graph)`)

jamtur01:
@mcorbin Any thoughts?

mcorbin:
Hi,

I completely refactored the influxdb stream in Riemann 0.2.13. Before 0.2.13, Riemann used its own method to construct the influxdb messages. Now, Riemann uses the official influxdb java client.

I was able to reproduce the first error (but inverted, float instead of int).

Riemann uses [this function](https://github.com/influxdata/influxdb-java/blob/master/src/main/java/org/influxdb/dto/Point.java#L125) (now deprecated btw) to construct the Influxdb Point object.

As you can see, the value is always converted to double.

[...]

PR #849, opened 2017-08-29
Converts clojure.lang.BigInt to double in influx stream

mcorbin:
The influxDB Java client does not support clojure.lang.BigInt (cf #848 ).


jamtur01:
LGTM

PR #854, opened 2017-09-19
Add a try/catch where events are expired

### LINKED ISSUE #850: Catch exceptions during event expiration
If you have a malformed event in the index (for example an event with a string :ttl), events will never expire because an exception will occur [here](https://github.com/riemann/riemann/blob/master/src/riemann/index.clj#L77).

I think we should catch the exception in the filter fn.

mcorbin:
Fix #850 

If a malformed event is indexed, and cause an exception during its expiration, the events in the index will never expire (the reaper thread is killed).

This commit adds a try/catch to catch the exception, log an error message, and remove the malformed event from the index.

### DIFF
--- a/src/riemann/index.clj
+++ b/src/riemann/index.clj
@@ -13,7 +13,8 @@
   (:refer-clojure :exclude [update])
   (:require [riemann.query :as query]
             [riemann.common :refer [deprecated]]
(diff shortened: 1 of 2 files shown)


jamtur01:
Thanks!

PR #861, opened 2017-10-03
Refactoring the telegram output

mcorbin:
- Breaking change: refactoring stream options. 

Before: 

```clojure
(streams
   (telegram {:token token
              :chat_id chat_id}))
```

Now : 

```clojure
(streams
    (telegram {:token token
               :telegram-options {:chat_id chat-id}}))
```

The `:telegram-options` key can contains all telegram [sendMessage](https://core.telegram.org/bots/api#sendmessage) options.

- New option: `:http-options` for http options like proxy
- New option: `:message-formatter` a function accepting an event and returning a string.
- I also simplified the code and added some tests.

I need to test this PR using a telegram account before merging ;)

### DIFF
--- a/src/riemann/telegram.clj
+++ b/src/riemann/telegram.clj
@@ -5,77 +5,78 @@
 
 (def ^:private api-url "https://api.telegram.org/bot%s/%s")
 
-(defn- html-parse-mode []
+(defn html-parse-mode
   "Formats html message."
-  (fn [e]
-    (str
-      "<strong>Host:</strong> " (or (:host e) "-") "\n"
-      "<strong>Service:</strong> " (or (:service e) "-") "\n"
-      "<strong>State:</strong> " (or (:state e) "-") "\n"
-      "<strong>Metric:</strong> " (or (:metric e) "-") "\n"
-      "<strong>Description:</strong> " (or (:description e) "-"))))
+  [e]
+  (str
+   "<strong>Host:</strong> " (or (:host e) "-") "\n"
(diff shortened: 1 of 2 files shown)


src/riemann/telegram.clj
```diff
@@ -5,77 +5,78 @@
...
+  Format event (or events) to string with markdown syntax by default.
 
   Telegram bots API documentation: https://core.telegram.org/bots/api
 
+  Options:
+
+  `:token`              The telegram token
+  `:http-options`       clj-http extra options (optional)
+  `:telegram-options`   These options are merged with the `:form-params` key of
+  the request. The `:chat_id` key is mandatody.
```
jamtur01: mandatory

src/riemann/telegram.clj
```diff
@@ -5,77 +5,78 @@
...
+  By default, the `:parse_mode` key is \"Markdown\".
+  `:message-formatter`  A function accepting an event and returning a string. (optional).
+  If not specified, `html-parse-mode` or `markdown-parse-mode` will be used,
+  depending of the `:parse_mode` value.
```
jamtur01: depending on

src/riemann/telegram.clj
mcorbin: done, thanks.

src/riemann/telegram.clj
mcorbin: done, thanks.

jamtur01:
Awesome!

PR #862, opened 2017-10-10
Allows to generate test results in junit format

mcorbin:
I wanted the test results in junit format.

I added 2 new options, `test.output.format=junit` to have the test output in JUnit format, and `test.output.file` to save the test results in a file.

For example:

```clojure
(tests
 (deftest foo
   (is (= 1 2))))
```
```
riemann -Dtest.output.format=junit -Dtest.output.file=/tmp/foo test /etc/riemann/riemann.config
```

```
cat /tmp/foo 
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
    <testsuite name="config-test" package="riemann">
        <testcase name="foo" classname="riemann.config-test">
            <failure>expected: 1
  actual: (2)
      at: MultiFn.java:229</failure>
        </testcase>
    </testsuite>
</testsuites>
```
Unfortunaly, the failure description seems to always be "MultiFn.java".

### DIFF
--- a/src/riemann/bin.clj
+++ b/src/riemann/bin.clj
@@ -5,6 +5,7 @@
             riemann.time
             [riemann.test :as test]
             riemann.pubsub
+            [clojure.test.junit :refer [with-junit-output]]
             [cemerick.pomegranate :as pom]
             [clojure.java.io :as io]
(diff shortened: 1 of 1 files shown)


jamtur01:
Thanks!

Issue #838, opened 2017-07-24
Add index support in test mode


mcorbin:
Hi,

I want to implement the possibility to test/query the index in test mode. Actually, streams like [maintenance-mode](http://riemann.io/howto.html#query-the-index-from-within-a-stream) are not testable.

My last attempt to do it was https://github.com/riemann/riemann/pull/821, and i have tried to do something better but i have trouble doing it.

To test the index we need `riemann.config/core` defined, because the index stream update the index in this core. Actually, only riemann.config/next-core is defined (because in test mode `riemann.config/apply!` is not called).
Furthermore, we cannot import riemann.config in riemann.test because of cyclic dependency.

I tried that:

In test.clj, i defined a *test-core* variable instead of *streams*

```
(def ^:dynamic *test-core*
  nil)
```

Now, in bin.clj, i added:

```clojure
(reset! riemann.config/core @riemann.config/next-core)
;; bind *test-core* instead of *streams*
(binding [test/*test-core* @config/core]
   ...
)
```

Now, i have riemann.config/core and riemann.test/\*test-core\* defined, and i can use \*test-core\* in test.clj (for example in inject! with `(inject! (:streams *test-core*) events)`).
I can also clear the index between each tests, and can query the index in tests.

I also tried to add expiration support in test mode. The reaper uses `(Thread/sleep interval)`, so it won't work.

My first tought was to define a `(periodically-expire interval)` function in test.clj. Once called (in a test for example), it will schedule using `riemann.time/every!` a sort of reaper, expiring events.
Unfortunately, inject! call `(time.controlled/reset-time!)` so my scheduled task is cancelled.

And this is where i am, and i'm a bit stuck, and will really appreciate help/guidance. I think Riemann really need this feature.

mcorbin:
So, i maybe have a solution, i hope it will be good enough. First, i created a \*test-core\* variable like i explained before.
Like i said, inject! call `(time.controlled/reset-time!)` so i cannot schedule a "reaper" outside of it.

So i defined a new variable in test.clj (yeah more state...) :

```clojure
(def ^:dynamic *test-config*
  {})
```
This variable will contains the configuration for each test (example soon).

Now, i added in inject!, after reset-time! these lines:

```clojure
(when-let [periodically-expire-opts (:periodically-expire *test-config*)]
         (periodically-expire-test periodically-expire-opts streams))
```

[...]

PR #871, opened 2017-11-09
Mock a core in test mode, refactoring reaper

mcorbin:
cf #838 for PR description

- Mock a core in test mode
- The reaper is now a task and not a service, so it will run on the Riemann scheduler.
- Also replaced "aphyr.com" by "127.0.0.1" in influxdb tests, because dns resolution breaks the test suite if offline mode ;)

This PR needs more tests/reviews, but i wanted to show the code now to get feedback on it.

### DIFF
--- a/src/riemann/kafka.clj
+++ b/src/riemann/kafka.clj
@@ -1,7 +1,8 @@
 (ns riemann.kafka
   "Receives events from and forwards events to Kafka."
   (:require [kinsky.client :as client]
-            [cheshire.core :as json])
+            [cheshire.core :as json]
+            [riemann.test :as test])
   (:use [riemann.common        :only [event]]
         [riemann.core          :only [stream!]]
         [riemann.service       :only [Service ServiceEquiv]]
@@ -103,8 +104,9 @@
       (conflict? [this other]
         (= opts (:opts other)))
       (start! [this]
-        (info "Starting kafka consumer")
-        (start-kafka-thread running? core opts))
+        (when-not test/*testing*
+          (do (info "Starting kafka consumer")
+              (start-kafka-thread running? core opts))))
       (reload! [this new-core]
         (info "Reload called, setting new core value")
         (reset! core new-core))
--- a/src/riemann/test.clj
+++ b/src/riemann/test.clj
@@ -7,6 +7,8 @@
   received."
   (:require [riemann.time.controlled :as time.controlled]
             [riemann.time :as time]
+            [riemann.index :as index]
+            [riemann.service :as service]
             [riemann.streams :as streams]
             [clojure.test :as test]))
 
@@ -24,10 +26,8 @@
   "An atom to a map of tap names to information about the taps; e.g. file and
   line number, for preventing collisions." nil)
 
-(def ^:dynamic *streams*
-  "The current sequence of streams. This is global so test writers can simply
-  call (inject! events) without looking up the streams from their core every
-  time."
+(def ^:dynamic *core*
+  "The core used in test mode"
   nil)
 
 (defn tap-stream
@@ -128,27 +128,22 @@
   riemann.time.controlled is global. Streams may be omitted, in which case
   inject! applies events to the *streams* dynamic var."
   ([events]
-   (inject! *streams* events))
+   (inject! (:streams *core*) events))
   ([streams events]
-   (binding [*results* (fresh-results @*taps*)]
-     ; Set up time
-     (time.controlled/with-controlled-time!
-       (time.controlled/reset-time!)
+   ;; Apply events
+   (doseq [e events]
+     (when-let [t (:time e)]
+       (time.controlled/advance! t))
 
-       ;; Apply events
-       (doseq [e events]
-         (when-let [t (:time e)]
-           (time.controlled/advance! t))
+     (doseq [stream streams]
+       (stream e)))
+     ;; Return captured events
 
-         (doseq [stream streams]
-           (stream e)))
-
-       ;; Return captured events
-       (->> *results*
-            (reduce (fn [results [tap-name results-atom]]
-                      (assoc! results tap-name @results-atom))
-                    (transient {}))
-            persistent!)))))
+   (->> *results*
+        (reduce (fn [results [tap-name results-atom]]
+                  (assoc! results tap-name @results-atom))
+                (transient {}))
+        persistent!)))
 
 (defn lookup
   "Lookup an event by host/service in a vector of tapped events returned by
@@ -173,10 +168,14 @@
   [name & body]
   `(test/deftest ~name
      (binding [*results* (fresh-results @*taps*)]
-       (time.controlled/with-controlled-time!
-         (time.controlled/reset-time!)
-
-         ~@body))))
+    (when (:index *core*)
+      (index/clear (:index *core*)))
+    (time.controlled/with-controlled-time!
+      (time.controlled/reset-time!)
+      (dorun (pmap #(riemann.service/reload! % *core*) (:services *core*)))
+      (dorun (pmap service/start! (:services *core*)))
+      ~@body
+      (dorun (pmap service/stop! (:services *core*)))))))
 
 (defmacro tests
   "Declares a new namespace named [ns]-test, requires some clojure.test and
@@ -201,7 +200,7 @@
   `(let [out# (atom [])
          stream# (~@stream (streams/append out#))]
      (time.controlled/reset-time!)
-     (doseq [e# ~inputs] 
+     (doseq [e# ~inputs]
        (when-let [t# (:time e#)]
          (time.controlled/advance! t#))
        (stream# e#))
--- a/test/riemann/core_test.clj
+++ b/test/riemann/core_test.clj
@@ -1,6 +1,7 @@
 (ns riemann.core-test
   (:require [riemann.client :refer :all]
             [riemann.common :refer [event]]
+            [riemann.config :as config]
             [riemann.core :refer :all]
             [riemann.index :refer [index]]
             [riemann.logging :as logging]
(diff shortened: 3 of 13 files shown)


aphyr:
This PR seems to do three unrelated things, so I'd like you to break it up into three parts. Changing the influx test is a no-brainer; I'm not sure why they used aphyr.com anyway, since it doesn't actually run an influx server. Mocking a core... not sure yet. The reaper changes I don't really understand--it seems like you've completely gutted the service machinery when you could have the service call riemann.time for scheduling instead--note that riemann.time/cancel can be used to cancel scheduled tasks. That'd save the trouble of a drastic rewrite, and then you can use the same logic to support all kinds of services, instead of having to special-case the reaper.

mcorbin:
Thanks you for your review @aphyr 

I just updated my PR. It's probably not perfect yet but i wanted to show you some code.

I created a new type if service, `EveryTaskService`, which schedule a task calling `(f core)` every interval seconds. I use it for the reaper.

So the reaper is now executed in the Riemann scheduler, and it is still a service. We just need to start it in inject!.

Regarding the influx test, the test was mocked but it did a dns lookup. I did another PR to fix it. 

If you want, i can divide this PR into 2 PR:

- One converting the reaper to an EveryTaskService
- One mocking a core in test mode.

Or i can do everything in this PR (and will rebase/reorganize commits).

mcorbin:
EveryTaskService renamed ScheduledTaskService ;)

mcorbin:
I updated the PR.

I modified the services (tcp, udp, websocket servers etc...) to not start in test mode (using the riemann.test/*testing* variable).

I now reload/start all services in `inject` before injecting events, and stop services after.

I will continue to do some tests, but everything seems to work. 
Is my design ok ?

src/riemann/kafka.clj
```diff
@@ -103,8 +104,9 @@
       (conflict? [this other]
         (= opts (:opts other)))
       (start! [this]
-        (info "Starting kafka consumer")
-        (start-kafka-thread running? core opts))
+        (if-not test/*testing*
```
aphyr: I think you need tests for this.

src/riemann/test.clj
```diff
@@ -24,10 +26,8 @@
   "An atom to a map of tap names to information about the taps; e.g. file and
   line number, for preventing collisions." nil)
 
-(def ^:dynamic *streams*
-  "The current sequence of streams. This is global so test writers can simply
-  call (inject! events) without looking up the streams from their core every
-  time."
+(def ^:dynamic *test-core*
+  "The core used in test mode"
```
aphyr: This namespace is already about testing; is there a reason to call this variable out as `test-core` instead of just `core`?

src/riemann/test.clj
```diff
@@ -173,6 +174,8 @@
   [name & body]
   `(test/deftest ~name
      (binding [*results* (fresh-results @*taps*)]
+       (if (:index *test-core*)
+         (index/clear (:index *test-core*)))
```
aphyr: I don't think it's a good idea to leak state out of a test; we should ensure cores are spun up and shut down at the beginning and end of the body, respectively.

test/riemann/core_test.clj
```diff
@@ -251,61 +252,53 @@
...
-           (advance! 2)
-           (Thread/sleep 100)
-           (is (= @res {:tags ["hi"]
-                        :time 2
-                        :state "expired"}))))
+  (let [index (wrap-index (index))
+        res (atom nil)
+        expired-stream (riemann.streams/expired
+                        (partial reset! res))
+        reaper (reaper 0.1 {:keep-keys [:tags]})
```
aphyr: Why the change to the reaper interval here?

test/riemann/core_test.clj
```diff
@@ -251,61 +252,53 @@
...
-           ; Check that expired-stream received them.
-           (is (= @res
-                  {:service 1
-                   :time 0.011
-                   :state "expired"}))))
+  (let [index (wrap-index (index))
+        res (atom nil)
+        expired-stream (riemann.streams/expired
+                        (fn [e] (reset! res e)))
+        reaper (reaper 0.1)
```
aphyr: Why is the reaper interval changed? Do test users also need to change their use of reaper?

src/riemann/kafka.clj
mcorbin: What a dumb mistake, sorry about that :/ I wrote tests for tcp/udp/ws/sse transports
I am not sure about how write a test for the kafka service, because it does not expose its state (unlike the tcp server when i can use the `:killer` keys in my test).

src/riemann/test.clj
mcorbin: Renamed.

src/riemann/test.clj
mcorbin: In this PR, i keep the same core during all tests, i just start/stop all services before/after injecting events (i am forced to do it in `inject!` because  `inject!` call `reset-tasks!` and kill the new reaper which is now a task). 

I could add index/clear in `inject!`, before and after starting and stopping the services. Is it ok for you ?

test/riemann/core_test.clj
mcorbin: I changed the reaper interval because the reaper was before executed in a threadservice with thread/sleep (which take milliseconds). Now the reaper is executed in the Riemann scheduler which take time in seconds. That's why this test was broken and why i rewrote it.

test/riemann/core_test.clj
mcorbin: cf previous answer ;)

test/riemann/core_test.clj
aphyr: So... this is also a breaking API change to `reaper`? How do you think users will feel about that?

aphyr:
In general, I advise using (when-not) instead of (if-not) for single-tail conditionals (e.g. side effects); it'll help you avoid this class of bug.

test/riemann/core_test.clj
mcorbin: No, no change for the user. before, (periodically-expire 10) will call thread/sleep (* 1000 10), now it will just schedule a task every 10 seconds. So no breaking change.

src/riemann/test.clj
aphyr: Urrgh, yeah, this brings up the awkward question of what the semantics should be for multiple calls to `inject!` in a single deftest. I can see arguments for resetting the core and scheduler in `deftest`, but that is a breaking change to `inject!`. Maybe best to have a `with-core` macro introducing taps, a fresh core, and time context, and to have `deftest` wrap that around the test body.

mcorbin:
Yes, i replaced if-not by when-not.

test/riemann/core_test.clj
aphyr: So if the API didn't change... why are you changing the tests?

mcorbin: I modified the test because the test was failing with (for example for the `expires` test):

```clojure
Fail in expires
expected: (= @res {:service 1, :time 0.011, :state "expired"})
  actual: (not
 (=
  {:service 1, :state "expired", :time 0.010000000000000002}
  {:service 1, :time 0.011, :state "expired"}))
```

I think 0.001 was too small for the Riemann scheduler, advance! was causing rounded values.

src/riemann/test.clj
mcorbin: Why is it a breaking change ? `inject!` already reset time to 0 and reset the Riemann tasks running on the scheduler.

It tried to not use the same core during tests and recreate a new one for each test, but there is a cyclic dependency between riemann.test and riemann.config so i am not sure how to achieve this.

aphyr: > I can see arguments for resetting the core and scheduler in deftest, but that is a breaking change to inject!

Moving the core/scheduler/tap setup out to deftest instead of inject! would change the semantics of inject!.

mcorbin: yes, understood ;)

mcorbin: > Maybe best to have a with-test-context macro introducing taps, a fresh core, and time context, and to have deftest wrap that around the test body.

I could do this (and also remove from `inject!` the `\*result\*` binding, the calls to `time.controlled` and the core restart), but as you said it will be a breaking change.

If the breaking change is ok for you, i will try to update the PR tomorrow.

mcorbin: I just pushed a commit when i moved all the code concerning controlled time and core reset from `inject!` to deftest (i haven't created a new macro yet, maybe moving the code in deftest is enough). It seems to work, but it's a breaking change

We should now decide, should we break inject! or not ? ;)

mcorbin:
Should i do something more ? Once we decide if we change the semantics of inject!, i will clean the PR (modify the inject! code or not, make more tests, merge commits together...).

mcorbin:
I made some minor modifications on this PR (docstring, replaced an if with a let). I think we should only reset the scheduler/time in `deftest` and not in `inject!` (that's why my code does).
I will make more tests and rebase my commits soon, and after that i think my PR should be OK to merge.

aphyr:
Yeah, I think that's a good call, @mcorbin. Looking forward to merging. :)

mcorbin:
I wrote some configs and tests:

- multiple call to inject! does not reset the time anymore:

```clojure
(streams
 (fixed-time-window 5 (tap :test-tap)))

(tests
 (deftest foo-test
   (let [result (inject!
                 [{:time 1}
                  {:time 4}
                  {:time 5}
                  {:time 6}
                  {:time 11}
                  ])]
     (is (= (:test-tap result)
            [[{:time 1}
              {:time 4}
              {:time 5}]
             [{:time 6}]])))
   (let [result (inject! [{:time 21}])]
     (is (= (:test-tap result)
            [[{:time 1}
              {:time 4}
              {:time 5}]
             [{:time 6}]
             [{:time 11}]
             []])))))
```

- index expiration :

```clojure
(periodically-expire 10)

(streams
 (where (and (not (service #"^riemann ")) (not (expired? event)))
   #(info %)
   (index))
 (where (expired? event)
  #(info %)
  (tap :expired)))

(tests
 (deftest test-expired
   (let [result (inject! [{:host "foo" :tags ["t1"] :time 0 :ttl 10}
                          {:host "bar" :time 15 :ttl 10}
                          {:host "baz" :time 30 :ttl 10}
                          {:host "foobar" :time 40 :ttl 10}])]
     (is (=(:expired result)
           [{:host "foo", :state "expired", :time 20}
            {:host "bar", :state "expired", :time 30}]))))
  (deftest test-expired2
    (let [result (inject! [{:host "foo" :tags ["t1"] :time 0 :ttl 10}
                           {:host "bar" :time 15 :ttl 10}
                           {:host "baz" :time 30 :ttl 10}
                           {:host "foobar" :time 40 :ttl 10}])]
      (is (=(:expired result)
            [{:host "foo", :state "expired", :time 20}
             {:host "bar", :state "expired", :time 30}])))))

```

reinject:

```clojure
(streams
 (where (not (service "reinject"))
        (with :service "reinject" reinject))
 (where (service "reinject")
        (tap :reinject)))

(tests
 (deftest reinject-test
   (let [result (inject! [{:service "foo"}])]
     (is (= (:reinject result)
            [{:service "reinject"}])))))
```

All tests are passing. I also tried querying the index in a test, it works (the code is in #838).

async queue:

```clojure
(streams
 (async-queue! :foo {:queue-size 1000
                     :core-pool-size 1
                     :max-pool-size 1}
               (tap :queue-tap)))

(tests
 (deftest test-queue
   (let [result (inject! [{:host "foo" :service "bar" :time 0}])]
     (Thread/sleep 1000)
     (is (= (:queue-tap
             result)
            [{:host "foo" :service "bar" :time 0}])))))
```

This test sometimes fails, i think it's because sometimes events are not yet consumed from the queue so the tap is empty. Users could wrap the async-queue! in (io) (as it must be done actually because sending events to a stopped async queue produces exception).

I also tried to think about other potential bugs/side effects, but I did not find anything. So i think we can merge ;)

aphyr:
It looks like there's some sort of a nondeterministic bug in expiry still? One of the tests didn't pass on one of the Travis runs.

mcorbin:
The failing test is not in expiry, the log in `lein test riemann.index-test` is normal (we want to test if an exception occurs if a malformed event in the index is expired). We should probably suppress this log in test mode.

The failing test is not `FAIL in (threadpool-service-test) (service_test.clj:155)`. The diff is:

```clojure
;; actual
{:service "riemann executor cat completed rate", :metric 2/5, :tags ["riemann"], :state "ok", :time 5}
{:service "riemann executor cat threads active", :metric 1, :tags ["riemann"], :state "ok", :time 5}
;; expected
{:service "riemann executor cat completed rate", :metric 3/5, :tags ["riemann"], :state "ok", :time 5}
{:service "riemann executor cat threads active", :metric 0, :tags ["riemann"], :state "ok", :time 5}
```

I don't think this is related to my PR, i didn't touch this part of the code.

aphyr:
Okay, let's go ahead and merge that then. I'm... real nervous about that failing test though; don't remember ever seeing that before. Can you take a look?

mcorbin:
Thanks for the merge ;) I created #894 for the failing test.

PR #873, opened 2017-11-11
Refactor test imports and fix indentation

mcorbin:
Sometimes Riemann uses `:uses` for import, sometimes `:require`, sometimes both. I think we should only use `:require` (as advised since clojure 1.4).

Also, Indentation is sometimes different even between differents functions in the same file, so it would be nice to reindent some functions.

This PR do it for the first 6 namespaces in riemann.test. If you are OK with that, i can do the PRs for every namespaces in Riemann (6 namespaces/PR to avoid big PR).

### DIFF
--- a/test/riemann/blueflood_test.clj
+++ b/test/riemann/blueflood_test.clj
@@ -1,15 +1,14 @@
 (ns riemann.blueflood-test
-  (:use riemann.blueflood
-        riemann.streams
-        riemann.time.controlled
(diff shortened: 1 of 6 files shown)


jamtur01:
Thanks. Seems reasonable to do this everywhere.

mcorbin:
Ok, i will continue to do it in the next few days/weeks. Thanks for the merge ;)

Issue #870, opened 2017-11-09
Lein deps fails with lein 2.8.1


AndreaCrotti:
I was having issues with one project that has Riemann as dependency where running `lein deps` would fail with this error:

```
Tried to use insecure HTTP repository without TLS.
This is almost certainly a mistake; however in rare cases where it's
intentional please see `lein help faq` for details.
```

I managed to isolate the problem to Riemann itself, since I can reproduce the same problem with a freshly checked out Riemann.
I don't see  see anything obviously wrong there but one of the dependencies clearly declares as source an HTTP server, which in newer version of Lein is not allowed.

jamtur01:
Which dependency is that?

AndreaCrotti:
I found it, it's actually this causing the issue:
https://github.com/bracki/clj-nsca/blob/master/project.clj

so `clj-nsca`
And it's a simple fix in theory, looks quite abandoned that project though

jamtur01:
Added https://github.com/bracki/clj-nsca/pull/1.

PR #875, opened 2017-11-21
Allow for not using `index-suffix`

derekchiang:
Currently, if `index-suffix` is set to `nil` or an empty string, you get an error because `time-format/formatter` won't accept it.

This patch makes it possible to not use an `index-suffix`.

### DIFF
--- a/src/riemann/elasticsearch.clj
+++ b/src/riemann/elasticsearch.clj
@@ -73,7 +73,9 @@
             es-endpoint (format "%s/%s%s/%s"
                                 (:es-endpoint opts)
                                 (:es-index opts)
-                                (time-format/unparse (time-format/formatter (:index-suffix opts)) (datetime-from-event event))
+                                (if (empty? (:index-suffix opts))
+                                  ""
+                                  (time-format/unparse (time-format/formatter (:index-suffix opts)) (datetime-from-event event)))
                                 (:type opts))
             http-options {}]
         (post


mcorbin:
@jamtur01 #870 seems to break travis :s We should maybe fork clj-nsca and do something. 

The problematic dependency in clj-nsca is `[com.googlecode/jsendnsca-core "1.3.1"]`, and is also available in https://maven.repository.redhat.com/ga/ (cf https://mvnrepository.com/repos/redhat-ga). 
The lib is also on clojars (https://clojars.org/com.googlecode.jsendnsca/jsendnsca) but not the 1.3.1 version.

We could also refactor clj-nsca to use the latest `jsendnsca` version (cf https://github.com/jsendnsca/jsendnsca), but it will be more work.

mcorbin:
Regarding the PR, is the tests passed, it's OK for me.

jamtur01:
Yeah. I've been meaning to do this. :(

mcorbin:
Don't hesitate to ping me if i can do something for help you about it.

mcorbin:
btw, https://github.com/bracki/clj-nsca is version `0.0.2` in project.clj but Riemann uses version `0.0.3`.

We could:

- Fork clj-nsca
- Rename the project to something else
- Uses `:repositories {"redhat" "https://maven.repository.redhat.com/ga/"}` in project.clj
- By the way, move the `[org.clojure/clojure "1.5.1"]` :dev profile and not in dependencies (and upgrade it to clojure 1.8.0), because we don't want to have the clojure 1.5 jar in Riemann
- Upload to clojars
- Modify Riemann to use the new dependency.

derekchiang:
Looks like CI is passing now.  Should we merge this?

jamtur01:
Yes - thanks.

PR #876, opened 2017-11-21
Add riemann-cond-dt plugin documentation

mcorbin:
Add riemann-cond-dt plugin doc.
I am not sure about the plugin description (`A plugin for verifying that a condition on an event is true during a time period.`). Is it understandable ?


jamtur01:
Neat!

boernd:
Can we move the plugins section to e.g. its own page? It's really hard to find...

PR #877, opened 2017-11-22
Add samplerr to plugins doc

boernd:
I hope @faxm0dem does not mind adding this to the docs because it rocks!


jamtur01:
:) Thanks!

PR #878, opened 2017-11-22
Use clj-nsca fork

### LINKED ISSUE #870: Lein deps fails with lein 2.8.1

mcorbin:
Fix #870 

I forked clj-nsca, renamed cljr-nsca, and uploaded it to clojars.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -40,7 +40,7 @@
     [riemann-clojure-client "0.4.5"]
     [less-awful-ssl "1.0.1"]
     [slingshot "0.12.2"]
-    [clj-nsca "0.0.3"]
+    [cljr-nsca "0.0.3"]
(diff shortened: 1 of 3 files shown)


jamtur01:
Oh wow! You rock sir!

PR #879, opened 2017-11-23
Updates to support Java9

jamtur01:
All the tests pass as such. Certainly a basis.

At some point, we'll need to do something about `less-awful-ssl` and its use of `javax.xml.bind`.


jamtur01:
cc @mcorbin

jamtur01:
I'm not going to merge this until Clojure 1.9 is GA but wanted eyes on prior to that.

mcorbin:
We should also modify the init scripts for .deb/rpm (to support both java 8 and java 9, probably by detecting the java version on the host).

jamtur01:
@mcorbin Let me know how that looks.

mcorbin:
The scripts seems ok to me. Regarding the java 9 jvm options, i don't know the JVM well enough to be able to say what are the best options.

PR #881, opened 2017-11-28
Allow for using Riemann as a dependency

derekchiang:
While using Riemann by writing a configuration file is perfectly fine
for most purposes, sometimes people (particularly Clojure developers)
might prefer to stick to their existing workflows that typically involve
a true project manager such as Leiningen or Boot.  These tools in turn
allow the users to make use of the dev tools they've come to depend on,
such as the REPL, linters, code formatters, etc.

This PR introduces a new entrypoint that is the "run" function.  Users
can then declare Riemann as a dependency, import the "run" function,
then simply start Riemann by calling the "run" function with a "setup"
function, which contains code that they would otherwise put into the
configuration file.


src/riemann/bin.clj
```diff
@@ -54,11 +54,11 @@
...
-      (sun.misc.Signal. "HUP")
-      (proxy [sun.misc.SignalHandler] []
-        (handle [sig]
-                (info "Caught SIGHUP, reloading")
-                (reload!))))))
+     (sun.misc.Signal. "HUP")
+     (proxy [sun.misc.SignalHandler] []
+       (handle [sig]
+         (info "Caught SIGHUP, reloading")
+         (reload!))))))
```
derekchiang: These are just some formatting changes.  I can remove them if you think they are distractions.

src/riemann/bin.clj
```diff
@@ -78,12 +78,11 @@
   "Process identifier, such as it is on the JVM. :-/"
   []
   (let [name (-> (java.lang.management.ManagementFactory/getRuntimeMXBean)
-               (.getName))]
+                 (.getName))]
     (try
```
derekchiang: Same as above.

derekchiang:
Could I get an opinion on this PR?

mcorbin:
What exactly is your workflow ? Why not just call (-main "start" "riemann.config") ?

derekchiang:
@mcorbin the problem is that since `riemann.config` is not a normal Clojure file that is part of a lein/boot project, it doesn't play nicely with nREPL-based dev tools such as LightTable, emacs-cider, etc.

The point of this PR is to make it possible to have all the code in a proper lein/boot project.

mcorbin:
Why not just call `(riemann.bin/-main "start" "/etc/riemann/riemann.config")` and `(riemann.bin/reload!)` in your repl ?

jeroenvandijk:
I'm also interested in using Riemann as a library (one stretch further) and add my own workflow and setup around it. Are there more people interested in this? I have already started a local fork to make changes that support this.

I have the feeling a core Riemann library without all the unnecessary plugin dependencies would be very helpful for this. The project would than have to be split up into two subprojects. The functionality of tools.deps could make this possible making much of an impact on the development workflow of Riemann itself. I'll continue locally and provide an example for this setup if there is enough interest.

shmish111:
I'm also interested in this, there are a few use cases I've come across such as testing that would be nicer if I could use the standard Clojure ecosystem.

jamtur01:
Thanks @derekchiang - apologies this took so long to merge!

PR #882, opened 2017-11-29
Add a not-expired stream

mcorbin:
Riemann has an `expired` stream, but not the opposite.

### DIFF
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -1295,9 +1295,18 @@
 (def tagged "Alias for tagged-all" tagged-all)
 
 (defn expired
-  "Passes on events with :state \"expired\"."
+  "Passes on expired events."
   [& children]
-  (apply match :state "expired" children))
+  (fn stream [event]
+    (when (expired? event)
+      (call-rescue event children))))
+
+(defn not-expired
+  "Passes on not expired events."
+  [& children]
+  (fn stream [event]
+    (when (not (expired? event))
+      (call-rescue event children))))
 
 (defn with
   "Constructs a copy of each incoming event with new values for the given keys,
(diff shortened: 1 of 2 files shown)


aphyr:
Maybe preferable to use `(where (expired? event))` and `(where (not (expired? event)) ...)`.

mcorbin:
`(where (expired? event))` will check the event `:state` and the event time, whereas the `expired` stream (and the `not-expired` one in my PR) only check the `:state`.

For example, with this example from the howto:

```clojure
(streams
  ; Email any expired event
  (expired (email "ops@foo.com"))

  ; Process only events which are *not* expired
  (where (not (expired? event))
    ...))
```
Events with `:state` != "expired" but with expired time will never be processed. It's rather confusing.

aphyr:
Yeah, we should probably update the `expired` stream to use `expired?` as well. It's a bit of a holdover from an earlier time.

mcorbin:
I updated the `expired` stream to use `expired?`. I also updated the `not-expired` stream. 
Filtering expired events is a common use case, so i think we should keep the `expired` stream and not deprecate it.

src/riemann/streams.clj
```diff
@@ -1295,9 +1295,18 @@
 (def tagged "Alias for tagged-all" tagged-all)
 
 (defn expired
-  "Passes on events with :state \"expired\"."
+  "Passes on expired events."
   [& children]
-  (apply match :state "expired" children))
+  (fn [event]
```
aphyr: Inner functions should have names for profiling; the convention here is `(fn stream [event])`.

mcorbin: I didn't know that, thanks you ;) updated.

jamtur01:
LGTM.

PR #885, opened 2017-12-14
Upgrade kinsky to 0.1.20

boernd:
Upgrades kafka-clients to 1.0.0. Tested with kafka 0.11.0.0 and 1.0.0

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -42,7 +42,7 @@
     [slingshot "0.12.2"]
     [cljr-nsca "0.0.3"]
     [amazonica "0.3.95" :exclusions [joda-time]]
-    [spootnik/kinsky "0.1.16"]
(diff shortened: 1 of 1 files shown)


mcorbin:
LGTM (but it will break users who use kafka < 0.11 no ?)

boernd:
I think due to the version mechanism it should also work with 0.10.x brokers: https://cwiki.apache.org/confluence/display/KAFKA/Compatibility+Matrix

"Now, newer versions of the Java clients and other clients that support KIP-35 such as librdkafka are able to fall back to older request types or throw appropriate errors if functionality is not available."

pyr:
It should be 0.1.20, not 0.1.19, I just cut a new release.

jamtur01:
Excellent!

PR #887, opened 2017-12-15
Refactor test imports

mcorbin:
- Stop using `:use` (which is sort of deprecated in Clojure) and replace it with `:require`
- `:require` imports are now sorted alphabetically (first Riemann imports, then other imports)
- Remove some unused imports.


jamtur01:
Awesome cleanup!

PR #889, opened 2017-12-16
Update project dependencies

mcorbin:
Update some dependencies.

I have not updated the com.cemerick/pomegranate dependency (it produces a bug, i will open an issue).


mcorbin:
Seems to break the travis build, i will try to fix it.

PR #900, opened 2018-01-09
Fix logsize-rotate

asonix:
The new logging backend is not exactly the same as it used to be. `.setMaxFileSize` takes a `FileSize.` now, not a string


mcorbin:
LGTM

jamtur01:
Thanks!

PR #908, opened 2018-02-09
Add missing atom deref to properly close kafka client

peffenberger:
Missing deref makes the condition always evaluate to true, thus never properly closing the kafka client.

To reproduce:

1. Create configuration with kafka client.
2. Change client's opts.
3. bin/reload!
4. Observe two kafka client threads, while there should be only one.


mcorbin:
Good catch ! LGTM

jamtur01:
Nice!

PR #909, opened 2018-02-13
Add support for tags in Graphite input and output

mcorbin:
Graphite now support tags (http://graphite.readthedocs.io/en/latest/tags.html, https://grafana.com/blog/2018/01/11/graphite-1.1-teaching-an-old-dog-new-tricks/).

This PR adds support for Graphite tags for the Riemann Graphite server and the Graphite output.

## Server

I added a new option `:tags` (default to false). If true, Graphite tags will be added as Riemann attributes. Example:

```clojure
(graphite-server {:host host :tags true})
```

If you send to the Riemann Graphite server this string:
`foo.bar;host=riemann.fr;rack=n1;environment=production 30 1518544178`

You will have in Riemann:

`Event{:host riemann.fr, :service foo.bar, :state nil, :description nil, :metric 30.0, :tags nil, :time 1518544178, :ttl nil, :rack n1, :environment production}`

## Client:

You can use the existing `:path` option in the Graphite client to convert Riemann attributes as Graphite tags.  If i want to converts the :host, :rack and :environment attributes into Graphite tags, i can use the `graphite-path-tags` function in `:path`.

```clojure
(def graph (graphite {:host "localhost"
                      :path (riemann.graphite/graphite-path-tags
                             [:host :rack :environment])}))
```
For example, the previous event will have a Graphite path equal to `foo.bar;host=riemann.fr;rack=n1;environment=production`.

### DIFF
--- a/src/riemann/graphite.clj
+++ b/src/riemann/graphite.clj
@@ -76,6 +76,26 @@
                                                 frac))))
       event)))
 
+(defn graphite-path-tags
+  "Returns a function which constructs a path for an event.
+  Takes the service with spaces converted to dots, followed by the tags
+  referenced in `tags` if they exists in the event.
+
+  Example:
+
+  (def graph (graphite {:path (graphite-path-tags [:host :rack])}))
+
+  {:host \"foo\" :service \"api req\" :rack \"n1\"}
+  will have this path: api.req;host=foo;rack=n1"
+  [tags]
+  (fn [event]
+    (let [service (replace (:service event) #" " ".")
+          tags (filter #(get event %) tags)
+          tag-string (join ";" (map #(str (name %) "=" (get event %)) tags))]
+      (if (= tag-string "")
+        service
+        (str service ";" tag-string)))))
+
 (defn graphite-metric
   "convert riemann metric value to graphite"
   [event]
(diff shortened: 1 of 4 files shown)


src/riemann/graphite.clj
```diff
@@ -76,6 +76,19 @@
                                                 frac))))
       event)))
 
+(defn graphite-path-tags
+  "Returns a function to construct a path for an event.
```
jamtur01: Returns a function which constructs a path...

src/riemann/graphite.clj
```diff
@@ -76,6 +76,19 @@
...
+  Takes the service with spaces converted to dots, followed by the tags
+  referenced in `tags` if they exists in the event."
```
jamtur01: I feel like maybe an example would help here?

src/riemann/graphite.clj
mcorbin: fixed, thanks ;)

src/riemann/graphite.clj
mcorbin: i added an example.

jamtur01:
Thanks!

PR #910, opened 2018-02-18
Add multiple events support to the Graphite stream

mcorbin:
Actually, you cannot send list of events (like the output of `batch`) to the Graphite stream.

I added a new fn on the Graphite protocol (`send-lines`) and implemented it for the Graphite TCP/UDP client. I also modified the Graphite stream to use it, so you can now send to the Graphite stream list of events.

### DIFF
--- a/src/riemann/graphite.clj
+++ b/src/riemann/graphite.clj
@@ -15,11 +15,13 @@
 
 (defprotocol GraphiteClient
   (open [client]
-        "Creates a Graphite client")
+    "Creates a Graphite client")
   (send-line [client line]
-        "Sends a formatted line to Graphite")
+    "Sends a formatted line to Graphite")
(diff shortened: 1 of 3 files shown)


jamtur01:
Yeah I think only the pickle receiver takes batches. 

Thanks!

mcorbin:
I continue to send line by line to Graphite, so it's not really batches from the graphite point of vue.

Before, you could not send so the (graphite) streams list of events. The events arrived one by one and for each event Riemann did : 
1) claim a connection from the Graphite connection pool
2) send the event to graphite
3) release the connection

Now, you can send batches of events. If the (graphite) stream receives a list/batch of event, he will:
1) claim a connection from the Graphite connection pool
2) send sequentially all the events to graphite
3) release the connection.

PR #912, opened 2018-02-22
config: lexicographically sort files before including them

pyr:
A common use of the include directory functionality is to
load files in order, allowing symbols referenced early
in the inclusion process in subsequently loaded files.

With a lexicographical sort, the canonical use-case
of using a numbered naming scheme for include files
is now possible.


pyr:
The build failure is not related to this PR and happens on only one of the JVM profiles

jamtur01:
Thanks @pyr !

PR #919, opened 2018-04-11
Fix incorrect assertions in tests

Mongey:
👋  I noticed some tests were only asserting the first arg, and as such would always pass.


jamtur01:
Thanks!

PR #922, opened 2018-04-25
Provide a Docker image

xrstf:
This PR adds a simple Dockerfile that can be used to build a ready-to-use Docker image for Riemann. After running

    docker build -t riemann .

the user ends up with an image that can be run via

    docker run --rm riemann

I don't know why, but using Java 9 or 10 results in the following warnings being printed when starting the container:

    WARNING: An illegal reflective access operation has occurred
    WARNING: Illegal reflective access by io.netty.util.internal.ReflectionUtil (file:/usr/lib/riemann.jar) to constructor java.nio.DirectByteBuffer(long,int)
    WARNING: Please consider reporting this to the maintainers of io.netty.util.internal.ReflectionUtil
    WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
    WARNING: All illegal access operations will be denied in a future release

If possible, I would very much appreciate an official image on Docker Hub, but this involves setting up an account and the neccessary Github integration, which is out of scope for this PR.

Concerning the `MAINTAINER`, I didn't know which name to put in there. I'm open to suggestions :-)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM clojure:lein AS builder
+
+WORKDIR /riemann
+ADD . .
+
+RUN lein uberjar && \
+    mv target/riemann-*-standalone.jar target/riemann.jar
+
+FROM openjdk:10.0-jre
+MAINTAINER Christoph Mewes <git@xrstf.de>
+
+EXPOSE 5555/tcp 5555/udp 5556
+CMD ["/bin/riemann", "/etc/riemann.config"]
+
+COPY --from=builder /riemann/pkg/tar/riemann.config /etc/riemann.config
+COPY --from=builder /riemann/pkg/tar/riemann /bin/riemann
+COPY --from=builder /riemann/target/riemann.jar /lib/riemann.jar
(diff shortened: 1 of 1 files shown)


jamtur01:
For MAINTAINER use:

```
MAINTAINER james+riemann@lovedthanlost.net
```

Worth adding a `ENTRYPOINT` or `CMD` for the second image?

jamtur01:
Oh and I have an org on Docker Hub - riemannio - that I've never used but happy to add an official image.

xrstf:
> Worth adding an ENTRYPOINT or CMD for the second image?

The second image has a command: `CMD ["/bin/riemann", "/etc/riemann.config"]` -- I didn't set it as the last step in the Dockerfile because it will probably change rarely, so that the layer can be cached.

I personally would use `/bin/riemann` as the entrypoint and `/etc/riemann.config` as the command, but that's just because I like my images to behave like "programs". But most of the images I've seen out there prefer the CMD style, making it easier to circumvent the default program and run something else (like a shell).

Whatever you like better, I suppose ;-)

jamtur01:
Oh - eyes dropped to the bottom of the Dockerfile and missed the `CMD`. No coffee yet.

I also prefer the `ENTRYPOINT` and `CMD` combination but not many others do. Happy to leave it as is.

jamtur01:
Thanks!

PR #923, opened 2018-04-25
Docker improvements

xrstf:
I forgot two things in my initial MR:

* The default configuration should listen on all network interfaces. This makes testing Riemann easier. It would be even easier if the stock config would have a `prn` in addition to the `index`, but I don't know what trouble that might cause of the other package formats.
* Using the slim OpenJDK image should save some disk space. Riemann probably never needs fancy UI libraries to work.


jamtur01:
Looks good.

PR #925, opened 2018-05-03
Support for Opsgenie v2 API

### LINKED ISSUE #918: Opsgenie API v2
Hello,

The opsgenie api v1 will be shut down on June 30, 2018 as it's said at https://docs.opsgenie.com/docs/migration-guide-for-alert-rest-api
Are you aware of that ?

Thank you

mcorbin:
Fix #918 

The alert creation call in the new API can be configured in multiple ways (https://docs.opsgenie.com/docs/alert-api#section-create-alert), so i added the possibility to customize the request body by passing a function generating the body to the stream.

This is a breaking change. I will add documentation about the opsgenie stream once the next Riemann version is released.

### DIFF
--- a/src/riemann/opsgenie.clj
+++ b/src/riemann/opsgenie.clj
@@ -3,21 +3,22 @@
   (:require [clj-http.client :as client])
   (:require [cheshire.core :as json]))
 
-(def ^:private alerts-url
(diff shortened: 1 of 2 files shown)


jamtur01:
Thanks!

PR #938, opened 2019-01-16
Add Zabbix support

vortura:
These changes add support for forwarding events to a Zabbix server using the [Zabbix Sender protocol](https://www.zabbix.com/documentation/4.0/manual/appendix/items/trapper). The `:service` field of the Riemann event is used as the key for the Zabbix item, with the `:metric` field used as the value.

I'm happy with the basic functionality, but I will confess that this is the first real piece of Clojure I have written. Any suggestions on improving the code would be gladly incorporated.


vortura:
Oh dear, that is embarrassing. You are correct of course. I updated a bunch of the docstrings before submitting the PR and I guess I must have mangled that argument then. Still getting to grips with paredit and parinfer.

I have updated the pull request to fix the error.

src/riemann/zabbix.clj
```diff
@@ -0,0 +1,116 @@
...
+  used as the value. Use:
+
+  (zabbix {:host \"localhost\" :port 10051})
+
+  Options:
+
+  - :host           Hostname of the Zabbix server or proxy to send events to.
+                    Default: localhost.
+  - :port           Port of the Zabbix host. Default 10051.
+  - :error-handler  Handler function for any exceptions (typically connection
```
mcorbin: For that, Riemann has `exception-stream`, so an error handler is not needed.

PR #940, opened 2019-01-31
add stream untag (inverse of tag)

deoqc:
Simple functionality missing from `streams`, ability to `untag` a event.


jamtur01:
Thanks!

PR #947, opened 2019-05-21
Modernize rpm packaging

smortex:
`lein fatrpm` does not produce a working package out-of-the-box on a recent CentOS 7 system.

This PR update the package dependencies and rely on systemd to manage the riemann service.  The generated rpm can be installed flawlessly on CentOS 7 and is running in production \o/


jamtur01:
Thanks! This is awesome. I've been meaning to do this for a while.

faxm0dem:
Hmmm this will break EL6, won't it?

jamtur01:
@faxm0dem The JRE package upgrade or systemd?

smortex:
Oh!  I have to admit that we do not use CentOS 6 and I completely forgot about this one.  maybe it's worth building two RPM packages, one targeted to CentOS 6 and the second targeted to CentOS 7?

jamtur01:
@smortex Sure. That seems reasonable.

faxm0dem:
I would be perfectly happy with either a switch `lein fatrpm --with-systemd` or a different subcommane `lein fatrpm_systemd` whatever's possible :-)

jamtur01:
I've addressed this - the packaging system will now create two rpms - one for EL6 and one for EL7+

Issue #950, opened 2019-07-19
Receiving exception on reload: java.lang.IllegalStateException: Could not find a suitable classloader to modify from clojure.lang.LazySeq@aa452d23


198d:
Recently upgraded to v0.3.3 and receive the following exception when trying to reload our config. We are using `pomegranate/add-dependencies` in the config. Riemann starts fine, but pukes on the `add-dependencies` call on reload. Have seen related issues #914 and #928, but don't see a clear solution. Attempted to use `config/depends` instead of `add-dependencies` directly which got rid of the errors on reload, but then I couldn't start Riemann. Appears `config/depends` is meant for "plugins" and complained about not finding some metadata file, if I recall.

```
INFO [2019-07-19 17:42:30,104] SIGHUP handler - riemann.bin - Caught SIGHUP, reloading
INFO [2019-07-19 17:42:30,107] SIGHUP handler - riemann.pubsub - Sweeping transient subscriptions.
ERROR [2019-07-19 17:42:30,264] SIGHUP handler - riemann.bin - Couldn't reload:
clojure.lang.Compiler$CompilerException: java.lang.IllegalStateException: Could not find a suitable classloader to modify from clojure.lang.LazySeq@aa452d23, compiling:(/etc/riemann/riemann.config:37:1)
        at clojure.lang.Compiler.load(Compiler.java:7526)
        at clojure.lang.Compiler.loadFile(Compiler.java:7452)
        at clojure.lang.RT$3.invoke(RT.java:325)
        at riemann.config$include.invokeStatic(config.clj:450)
        at riemann.config$include.invoke(config.clj:427)
        at riemann.bin$reload_BANG_.invokeStatic(bin.clj:46)
        at riemann.bin$reload_BANG_.invoke(bin.clj:36)
        at riemann.bin$handle_signals$fn__14141.invoke(bin.clj:62)
        at riemann.bin.proxy$java.lang.Object$SignalHandler$d8c00ec7.handle(Unknown Source)
        at sun.misc.Signal$1.run(Signal.java:212)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.IllegalStateException: Could not find a suitable classloader to modify from clojure.lang.LazySeq@aa452d23
        at cemerick.pomegranate$add_classpath.invokeStatic(pomegranate.clj:54)
        at cemerick.pomegranate$add_classpath.invoke(pomegranate.clj:42)
        at clojure.lang.AFn.applyToHelper(AFn.java:154)
[...]

PR #951, opened 2019-08-05
riemann.bin: Refactor ensure-dynamic-classloader and call on reload

198d:
Since upgrading Pomegranate to 1.0, issues have arose around the
classloader hierarchy that have triggered weird bugs around the use of
`add-dependencies` and strange behavior when handling the HUP signal.
This version of `ensure-dynamic-classloader` comes from studying how the
compiler and runtime find the approrpriate classloader and takes into
account differences across JDK versions in the classloader hierarchy
associated with the thread handling the HUP signal to ensure consistent
behavior.

I have tested these changes across OpenJDK 8, 9 and 10 using the `library/openjdk` container images found on DockerHub. See comments inline for a little more in-depth discussion about why I made some of the changes I made. This resolves many of the issues discussed in #950.

### DIFF
--- a/src/riemann/bin.clj
+++ b/src/riemann/bin.clj
@@ -50,6 +50,14 @@
         (error e "Couldn't reload:")
         e))))
 
+(defn ensure-dynamic-classloader
+  []
+  (let [thread (Thread/currentThread)
+        cl (or (.getContextClassLoader thread)
+               (.getClassLoader clojure.lang.Compiler))]
+    (when-not (instance? DynamicClassLoader cl)
+      (.setContextClassLoader thread (DynamicClassLoader. cl)))))
+
 (defn handle-signals
   "Sets up POSIX signal handlers."
   []
@@ -59,6 +67,7 @@
      (proxy [sun.misc.SignalHandler] []
        (handle [sig]
          (info "Caught SIGHUP, reloading")
+         (ensure-dynamic-classloader)
          (reload!))))))
 
 (defn pom-version
@@ -104,14 +113,6 @@
       (binding [clojure.test/*test-out* w]
         (run-tests-with-format test-name-pattern)))
     (run-tests-with-format test-name-pattern)))
-
-(defn ensure-dynamic-classloader
-  []
-  (let [thread (Thread/currentThread)
-        cl (.getContextClassLoader thread)]
-    (when-not (instance? DynamicClassLoader cl)
-      (.bindRoot clojure.lang.Compiler/LOADER cl)
-      (.setContextClassLoader thread (DynamicClassLoader. cl)))))
 
 (defn run-app!
   "Start Riemann with the given setup function. The setup function is responsible
--- a/src/riemann/bin.clj
+++ b/src/riemann/bin.clj
@@ -53,10 +53,12 @@
 (defn ensure-dynamic-classloader
   []
   (let [thread (Thread/currentThread)
-        cl (or (.getContextClassLoader thread)
-               (.getClassLoader clojure.lang.Compiler))]
-    (when-not (instance? DynamicClassLoader cl)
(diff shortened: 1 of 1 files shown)


src/riemann/bin.clj
```diff
@@ -59,6 +67,7 @@
      (proxy [sun.misc.SignalHandler] []
        (handle [sig]
          (info "Caught SIGHUP, reloading")
+         (ensure-dynamic-classloader)
```
198d: The classloader returned by `(.getContextClassLoader (Thread/currentThread))`  in the Thread handling the signal returns, under OpenJDK 8, the class loader hierarchy returned at the beginning of the program, not seeing the the hierarchy installed by `ensure-dyanmic-classloader` when Riemann starts; under OpenJDK 9 and 10 it returns `nil`, explaining the issues `0.3.3` attempted to fix by binding the class loader to `clojure.lang.Compiler/LOADER`. Calling our new `ensure-dynamic-classloader` function here ensures an appropriate class loader hierarchy for the config reload.

src/riemann/bin.clj
```diff
@@ -50,6 +50,14 @@
         (error e "Couldn't reload:")
         e))))
 
+(defn ensure-dynamic-classloader
+  []
+  (let [thread (Thread/currentThread)
+        cl (or (.getContextClassLoader thread)
+               (.getClassLoader clojure.lang.Compiler))]
```
198d: This aligns with the fallback in `clojure.lang.RT/baseLoader` as can be seen here in Clojure 1.9.0: https://github.com/clojure/clojure/blob/clojure-1.9.0/src/jvm/clojure/lang/RT.java#L2171.

src/riemann/bin.clj
198d: Unfortunately,  as I typed the above comment, I realized there's still one last bug that needs attention: the call to `riemann.bin/set-config-file!` uses `pomegranate/add-classpath` to configure the config file's parent directory on the classpath and will need to be called (or have that part factored out and called) to carry that configuration on during the reload since we essentially lose everything configured on startup with every reload.

src/riemann/bin.clj
198d: The above is necessary because of the behavior described in my other comment concerning the class loader hierarchy present in the Thread handling the HUP signal.

src/riemann/bin.clj
```diff
@@ -16,6 +16,15 @@
   "The configuration file loaded by the bin tool"
   (promise))
 
+(defn add-config-dir-to-classpath
+  [config-file]
+  (let [dir (-> config-file
+                io/file
+                .getCanonicalPath
```
greglook: You can avoid the extra `io/file` call below by using:
```suggestion
                .getCanonicalFile
```

jamtur01:
@198d I'm broadly happy assuming we can address @greglook's feedback. I was hoping for eyes on from @pyr but I think this works.

jamtur01:
Hey @198d just quickly following up to see if you might have a chance to finish this off? Thanks!

198d:
@jamtur01 Will try and get this tidied up this week. Thanks!

src/riemann/bin.clj
198d: `.getCanonicalPath` returns a path that is absolute. Removed `io/file` and simplified to `.getCanonicalFile`.

jamtur01:
@198d Any thoughts on the `.getAbsoluteFile` call?

198d:
@jamtur01 Hidden above by the resolved convo, but `getCanonicalFile` returns a File object with an absolute path according the JDK docs.

jamtur01:
Cool! Thanks @198d. Merging.

PR #953, opened 2019-08-22
Add predict-linear stream

boernd:
Builds a linear regression model to predict future `:metric` values (similiar to the `predict_linear` function in Prometheus). Can be used e.g. to predict filesystem or tablespace usage (no hard thresholds).


jamtur01:
Looks good. Thanks!

PR #955, opened 2019-09-01
Add syntax highlighting to datadog example in docs

rwaweber:
Not sure how to generate the docs html from comments, but I largely just followed the way that the druid example looked with the `clojure` backticks. I'm _assuming_ this will make the below example a smidge more readable on the docs site.


jamtur01:
Looks good. Thanks! The API docs are auto-generated when we release.

PR #965, opened 2020-01-08
fix(transport): service/conflict? doesn't fail if rabbitmq-transport isn't started

serge-medvedev:


src/riemann/transport/rabbitmq.clj
```diff
@@ -34,18 +34,24 @@
...
     (and (instance? RabbitMQTransport other)
-         (= (.getId @connection)
-            (-> other :connection (deref) (.getId)))))
+         (same-settings? settings (:settings other))))
  
   Service
   (conflict? [this other]
     (and (instance? RabbitMQTransport other)
-         (= (.getId @connection)
-            (-> other :connection (deref) (.getId)))))
```
serge-medvedev: This led to **NullPointerException** if I tried to create more than one instance of rabbitmq-transport in riemann.config.

serge-medvedev: ... now I get **IllegalArgumentException** which is expected behavior :)

PR #967, opened 2020-01-30
refactor(rabbitmq-transport): configuration made even more flexible

serge-medvedev:


faxm0dem:
Thanks for your changes. This will however break existing code. We should probably provide backwards-compatible changes

serge-medvedev:
I supposed that it's allowed to make any changes (even breaking ones) between releases - am I wrong?

faxm0dem:
From a semantic versioning perspective, you're right, as we're still in 0.x.y. I'd personally be in favour of breaking changes, but I'd prefer to have @aphyr @jamtur01 agree first

aphyr:
Yeah, Riemann has a large install base: compatibility and a deprecation notice would be my preference. Folks always appreciate it when you're considerate!--KyleOn Feb 3, 2020 11:54, Fabien Wernli <notifications@github.com> wrote:From a semantic versioning perspective, you're right, as we're still in 0.x.y. I'd personally be in favour of breaking changes, but I'd prefer to have @aphyr @jamtur01 agree first

jamtur01:
What @aphyr said. For next release I'd like backwards compatibility and a deprecation notice. x+1 releases beyond that we'd deprecate and change.

serge-medvedev:
I do agree with your reasoning, guys, but a bit of clarification is needed.
The latest release is 0.3.5 which **doesn't** contain rabbitmq-transport functionality in it, 0.3.6 isn't out there yet so what exactly changes in this PR has to be backward compatible with?

jamtur01:
Yeah I just realized this isn't in 0.3.5. I thought it was - mea culpa. I think we can merge without issues.

aphyr:
Oh, if that's the case, it should be fine!On Feb 3, 2020 16:00, Serge Medvedev <notifications@github.com> wrote:I do agree with your reasoning, guys, but a bit of clarification is needed.
The latest release is 0.3.5 which doesn't contain rabbitmq-transport functionality in it, 0.3.6 isn't out there yet so what exactly changes in this PR has to be backward compatible with?

faxm0dem:
haha same here, didn't realise this - sorry for the noise and waking everyone up

PR #968, opened 2020-02-06
feat(stream): rabbitmq stream is introduced

serge-medvedev:
It's useful if you want not only to receive events from RabbitMQ (via rabbitmq-transport) but send some, too.


jamtur01:
Thanks!

PR #971, opened 2020-04-16
Fix InfluxDB typo and improve copy

maddenp:
The code example in the `influxdb` section incorrectly uses the keyword `:user`, which should be `:username` (as seen in the _General Options_ section below the code example). Fix that, and tidy up some other commentary.


jamtur01:
Thanks!

Issue #941, opened 2019-02-11
inject! hangs up with current timestamps


bickerle:
Hello fellow riemann enthusiasts,

while running tests with current event timestamps I encountered the problem that my test events were not injected into the streams. The following snippet  showcases the issue. When run as is the test does not terminate. If you remove the last event (its timestamp has one more digit than the first two) it terminates within seconds.

```clojure
(tests
  (deftest timestamp-test
    (prn (inject! [{:host "test.example.com" :service "test" :metric 1337 :time 109889374}
                   {:host "test.example.com" :service "test" :metric 1337 :time 109889375}
                   {:host "test.example.com" :service "test" :metric 1337 :time 1109889374}]))))
```

I took a peak into the source code but could not find any reasons for this behavior.
The obvious workaround is to use smaller timestamps but it would be nice to be able to use real life examples as test cases.

Thanks in advance for any help!

mcorbin:
Hello,

I'm not able to reproduce your issue:

```
(logging/init {:file "/var/log/riemann/riemann.log"})

(let [host "127.0.0.1"]
  (tcp-server {:host host})
  (udp-server {:host host})
  (repl-server {:host host})
  (ws-server  {:host host}))

(streams
  (tap :foo))

(tests (deftest timestamp-test
         (let [result
               (inject! [{:host "test.example.com" :service "test" :metric 1337 :time 109889374}
                         {:host "test.example.com" :service "test" :metric 1337 :time 109889375}
                         {:host "test.example.com" :service "test" :metric 1337 :time 1109889374}])]
           (= (:foo result)
              [{:host "test.example.com" :service "test" :metric 1337 :time 109889374}
               {:host "test.example.com" :service "test" :metric 1337 :time 109889375}
               {:host "test.example.com" :service "test" :metric 1337 :time 1109889374}]))))
```
```
riemann test config.clj 
INFO [2019-02-11 19:30:55,013] main - riemann.bin - Loading /home/mathieu/Documents/config.clj
INFO [2019-02-11 19:30:55,073] main - riemann.repl - REPL server {:port 5557, :host 127.0.0.1} online

Testing riemann.config-test

Ran 1 tests containing 0 assertions.
0 failures, 0 errors.
```

What are you testing ?

bickerle:
Hi,

my apologies I should have added the whole config - but I managed to single out a possible culprit.
[...]

PR #974, opened 2020-04-27
Do not record events if *results* is nil.

sanel:
Behavior is explained on [1]

In short, unless (instrumentation) is explicitly set and tests has
time gap that will put Riemann in sleep, another thread will kick
instrumentation while tests are running, causing NullPointerException.

[1] https://github.com/riemann/riemann/issues/941#issuecomment-516172375


jamtur01:
Thanks!

PR #975, opened 2020-04-28
added pushover priorities support

mrkooll:
Added pushover priorities support. By default priority is based on riemann event state.

Also pushover notification time is set equal to riemann event time.


jamtur01:
Thanks!

development-lab:
I need code

PR #976, opened 2020-06-12
Pass http options to elasticsearch

smortex:
Allow the user to pass http-options to elasticsearch.

This is used for configuring a proxy (example stolen from another part of the elasticsearch module), but configuring persistent connections might be the best use-case…

Screenshot_2020-06-11 Riemann - Grafana


jamtur01:
LGTM! Thanks!

PR #985, opened 2020-10-26
Bump jackson-databind to mitigate vulnerabilities

sgerrand:
💁 A dependency of `logstash-logback-encoder`, `jackson-databind`, has numerous security vulnerabilities prior to version >= 2.10.x. Updating the version of `logstash-logback-encoder` and adding an explicit dependency on the right version of `jackson-databind` should mitigate this issue.

For further details on these CVS, please see: https://www.cvedetails.com/vulnerability-list.php?vendor_id=15866&product_id=42991


jamtur01:
Thanks!

PR #986, opened 2020-11-12
fix bash get JAVA_VERSION

x1e:
The openjdk-11-jdk:
```
java -version
openjdk version "11.0.9" 2020-10-20
```
The origin sed regex can not match, need fix.


ldziedziul:
Hi, we've hit the same problem. When are you going to release a version with this fix?

jamtur01:
@ldziedziul Maybe over the holiday season. I am afraid I'm pretty busy right now. Thanks for your patience.

PR #989, opened 2021-01-04
feat(stream): rabbitmq stream to use function or string as routing key

### LINKED ISSUE #988: rabbitmq routing-key as a function rather than a static string
**Is your feature request related to a problem? Please describe.**
The rabbitmq exchange routing-key is depending on an attribute on the event which is unable to be pre-defined.

**Describe the solution you'd like**
The routing-key for the exchange can be a function or string.

wimoMisterX:
To fix #988

### DIFF
--- a/src/riemann/rabbitmq.clj
+++ b/src/riemann/rabbitmq.clj
@@ -30,7 +30,7 @@
    :internal false}
   ```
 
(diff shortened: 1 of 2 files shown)


jamtur01 (approved):
LGTM

jamtur01:
Thanks!

Issue #993, opened 2021-03-11
Implement InfluxDB 2.0 plugin


bobkocisko:
Hello!  I am evaluating Riemann and realized that InfluxDB 2.0 doesn't appear to be supported yet as a destination to forward events.  Do you have any plans to implement it?

faxm0dem:
I suppose it would be pretty straightforward, unless they reworked the API too much

PR #996, opened 2021-05-16
InfluxDB v2 support

chhetripradeep:
Closes https://github.com/riemann/riemann/issues/993

I have tested it with my dev InfluxDB v2 environment.


jamtur01:
Awesome! Thanks.

Issue #957, opened 2019-10-01
config/index does not preserve index correctly


jarpy:
The `index` function in `config.clj` doesn't always preserve the existing index as expected. In particular, when `index` is called multiple times within the same config, the index is replaced, which results in odd behaviour.

Given this config:

```clojure
(info "Let binding index.")
(let [index (index)]
  (streams prn))

(info "Let binding index again.")
(let [index (index)]
  (streams prn))

(info "Calling (index).")
(index)
```

and adding some [debug logs to the index function](https://github.com/jarpy/riemann/commit/68929b5c01bb4afe06f7ab82d6408d80a9c02cd4), we see this output:
```
INFO [2019-10-01 22:25:22,834] main - riemann.bin - Loading /home/jarpy/src/riemann/riemann/repro.config
INFO [2019-10-01 22:25:22,838] main - riemann.bin - PID 5181
INFO [2019-10-01 22:25:22,840] main - riemann.config - Let binding index.
INFO [2019-10-01 22:25:22,843] main - riemann.config - ❌ Index replaced.
INFO [2019-10-01 22:25:22,844] main - riemann.config - Let binding index again.
INFO [2019-10-01 22:25:22,844] main - riemann.config - ❌ Index replaced.
INFO [2019-10-01 22:25:22,845] main - riemann.config - Calling (index).
INFO [2019-10-01 22:25:22,845] main - riemann.config - ❌ Index replaced.
INFO [2019-10-01 22:25:22,850] main - riemann.core - Hyperspace core online
```

This causes weird behaviour with the index in any config where `index` is called (probably let-bound) multiple times. In particular, expiry gets messed up. With this config:

```clojure
(instrumentation {:enabled? false})
(tcp-server  {:host "0.0.0.0"})
(periodically-expire 1)

(info "Let binding index.")
(let [index (index)]
  (streams
   (not-expired index prn)
   (expired prn)))

(info "Calling (index).")
(index)
```

when injecting an event with `:ttl 1`, we'd expect to print an expired event within 2 seconds, but it never arrives:

```
INFO [2019-10-01 22:29:17,943] main - riemann.bin - Loading /home/jarpy/src/riemann/riemann/repro.config
INFO [2019-10-01 22:29:17,948] main - riemann.bin - PID 6131
INFO [2019-10-01 22:29:17,969] main - riemann.config - Let binding index.
INFO [2019-10-01 22:29:17,973] main - riemann.config - ❌ Index replaced.
[...]

PR #997, opened 2021-07-02
Set current index on first call of config/index

jarpy:
Fix the "orphaned indices" problem described in #957.

Allows a Riemann config to make multiple calls to `riemann.config/index`
with predictable results, both on initial load and after a reload.

Without this change, multiple calls are OK after reload or explicit
apply, but create orphan indices on initial load.

Closes: #957

### DIFF
--- a/src/riemann/config.clj
+++ b/src/riemann/config.clj
@@ -246,7 +246,7 @@
   [& opts]
   (locking core
     (let [current-index (:index @core)
-          ; Create a new index, later we'll decide whether it needs to be added
+          ; Create a new index. Later we'll decide whether it needs to be added
           ; to the current core, next core, or both.
           ; Note that we need to wrap the *current* core's pubsub; the next
           ; core's pubsub module will be discarded in favor of the current one
--- a/src/riemann/config.clj
+++ b/src/riemann/config.clj
@@ -231,22 +231,39 @@
            (reduce conj (:streams @next-core) things))))
 
 (defn index
-  "Set the index used by this core. Returns the index."
+  "Create a new index and add it to the core and next core as needed.
+
+  If the current core has no index, add the new index to it.
+
+  If the next core has no index, add the current core's index (when equivalent)
+  or add the new index (when not equivalent).
+
+  In practice, because Riemann currently has only one index type (NBHM), only a
+  single index will ever be created. It will be preserved across all future
+  calls to `index`.
+
+  Returns the index."
   [& opts]
   (locking core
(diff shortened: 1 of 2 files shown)


jarpy:
I'd be chuffed to get some feedback on this proposal (after the current `log4j` emergency has calmed down, of course). Thanks.

jamtur01:
@faxm0dem @mcorbin Any thoughts?

faxm0dem:
Without looking at the implementation, my opinion is that this would be okay for 99% of use-cases. But what about the other 1%?
If we decide to merge this, I guess it would require a major version bump? Oh wait, we're still < 1.0...

jarpy:
I would like to do _something_ to make this easier for new users. I proposed a few solutions in #957, and maybe folks would be interested in exploring solutions in that issue. This PR is the solution that I prefer, because it doesn't add any new functions, or require more documentation to explain the subtleties of `index` (which should probably be called `index!`). That function has reasonably complex behaviour, and it wasn't obvious how it worked to me, when I was a new user (as we all were at some point). Even with extended documentation, it's something that people have to learn about and watch out for, instead of that something that just behaves predictably.

Even James' book, which is an exceptionally useful resource for new users, doesn't capture all the nuances. It reads that `index` is:

> [...] the function that sends events to Riemann's index. <sup>[1]</sup>.

In reality, `index` is

> _a function that returns the index attached to the core that will be used after the next core transition (triggered by config [re]load). As a side-effect, it may also replace that index with a brand new one, orphaning any reapers that are attached to the exiting (future) index. Whether this happens or not depends on global state. The index of the currently running core is unaffected._

That's a lot to reason about, and it trips people up. In particular, it trips up people who have their Riemann configuration spread over multiple files, if more than one of them, quite innocently, says:

```clojure
(let [index (index)])
```

That's a mutating call (sometimes), but how does the user know that? In practice, they don't know. They just end up with broken expiry, because they have reapers attached to indices that aren't attached to cores. Thankfully, my production config has a decent test suite, so it discovered that expiry stopped working, but I'm sure there are Riemann users who don't use the (amazing) test functionally, especially if they're new.

All in all, I'm arguing in favour of behaviour that is less subtle, and gives results that are more consistent when users do things that they see as examples in the documentation and The Art of Monitoring.

<sub>1. https://artofmonitoring.com/TheArtOfMonitoring_sample.pdf Page 23-24.</sub>

faxm0dem:
I pretty much agree to what you're saying, but I'll leave the decision to more experienced riemann contributors.

PR #998, opened 2021-07-14
Missing RabbitMQ docs are added

serge-medvedev:


serge-medvedev:
@jamtur01 , why is it not being merged? Do docs have to be updated in some other way I'm not aware of?

jamtur01:
@serge-medvedev No - more that I need to work out why it's not being auto-generated. I'll try to take a look this week.

jamtur01:
Okay. Fixed. I've merged the API docs now.

PR #999, opened 2021-09-18
Fix Linux detection for Netty

mcorbin:
Netty uses the epoll implementation only if Riemann detects
that the operating system is Linux based.

On my computer (Debian 10), `(System/getProperty "os.name")` returns
"Linux", which does not match the `#"linux"` regex.

This commit modifies the regex to also accept `Linux`.

Some REPL outputs:

(re-find  #"(linux|Linux)" (System/getProperty "os.name"))
["Linux" "Linux"]

(re-find  #"linux" (System/getProperty "os.name"))
nil


jamtur01:
Awesome. Thanks.

PR #1004, opened 2021-12-30
Add garbage collection timeout to `by` streams

jarpy:
Add a capability to `by` and `by-builder` that lets them expire streams, allowing the streams to be garbage collected.

Without this capability, declaring a `by` stream over a dimension that continuously changes will eventually exhaust all heap space. For example, `(by :host prn)`, when processing events from a dynamic fleet of VMs or containers, represents a fatal memory leak.

It's tempting to set the new behaviour as the default, with a sensibly long expiry time, but I'm sure it would break something for somebody, somewhere.

cc @mcorbin, because I believe that Mirabelle uses the same implementation.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -36,6 +36,7 @@
     [com.boundary/high-scale-lib "1.0.6"]
     [com.draines/postal "2.0.2"]
     [com.amazonaws/aws-java-sdk "1.11.116" :exclusions [joda-time]]
+    [expiring-map "0.1.9"]
     [interval-metrics "1.0.0"]
     [clj-antlr "0.2.4"]
     [io.netty/netty-all "4.1.41.Final"]
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -35,7 +35,8 @@
                              cancel]]
         clojure.math.numeric-tower
         clojure.tools.logging)
-  (:require [riemann.folds :as folds]
+  (:require [expiring-map.core :as exmap]
+            [riemann.folds :as folds]
             [riemann.index :as index]
             riemann.client
             riemann.logging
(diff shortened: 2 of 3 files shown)


jarpy:
I've been running this in production for a couple of weeks now, and it's been great.

mcorbin:
Hello,

You are in this PR if I'm not wrong garbage collecting the `by` stream but not the children in all cases.

For example, several Riemann streams schedule tasks in the internal task scheduler in order to produce a side effect every N seconds. Let's take this example:

```clojure
(by [:host]
  (coalesce 10
    (info)))
```

Every time we have a new `:host` value `coalesce` will be called. The coalesce stream will schedule a task which will be executed in my case every 10 seconds: https://github.com/riemann/riemann/blob/main/src/riemann/streams.clj#L1238

I think in some cases these tasks will never finish: https://github.com/riemann/riemann/blob/main/src/riemann/streams.clj#L523, and so even if `by` is garbage collected all tasks running in child streams will continue to run forever.

That's why I only use the events time to execute side effects in Mirabelle: to not have any global side effects (like schedulers) in streams and so to be able to remove streams as I want without fear (for hot reload, managing streams through the API or garbage collecting `by`, the last one being not implemented yet :D).

jarpy:
Very interesting. Thanks for the great explanation, Mathieu. I had not considered background tasks, just regular "realtime" streams.

I will probably continue to do something like this in my deployment, but I can see now that it is an incomplete solution to the memory leak problem.

mcorbin:
I think it's unfortunately not possible to implement a GC on `by` which will work in all cases, for all streams used as children.
if you make this that `the `:time` and the `:ttl` are always set on your events I think streams which use `periodically-until-expired` will be correctly garbage collected: https://github.com/riemann/riemann/blob/main/src/riemann/streams.clj#L521

A couple of streams are also calling `every!` directly and so will also create periodic tasks.

jarpy:
Cheers. I need to learn more about how tasks are managed.

In practice, I think that this patch is useful for my own environment. I _never_ allow events without `:time` and `:ttl` :D. Memory consumption is provably much, much better with my config. So, I think I'll implement this locally, because even if it's not perfect, it's pragmatic for my deployment.

PR #1014, opened 2022-05-14
Update docs for tagged-all, tagged-any

jarpy:
The docstrings for these streams had two issues:

- `tagged-any` contained an example of `tagged-all`
- Both suggested that the user can pass a string an an argument, but
  that will not work as expected.

Calling `(tagged-any "foo" prn)` will return a stream that prints events
that are tagged `(char 102)` or `(char 111)` (which are of type
`java.lang.Character`). It will not match the `java.lang.String` `"foo"`,
which is what the user would expect.

Refer: https://groups.google.com/g/riemann-users/c/KPnoqh8dC9M

### DIFF
--- a/src/riemann/streams.clj
+++ b/src/riemann/streams.clj
@@ -1294,7 +1294,6 @@
   Can be used as a predicate in a where form.
 
   ```clojure
-  (tagged-all \"foo\" prn)
   (tagged-all [\"foo\" \"bar\"] prn)
   ```"
   [tags & children]
@@ -1317,8 +1316,7 @@
   Can be used as a predicate in a where form.
 
   ```clojure
-  (tagged-any \"foo\" prn)
-  (tagged-all [\"foo\" \"bar\"] prn)
+  (tagged-any [\"foo\" \"bar\"] prn)
   ```"
   [tags & children]
   (let [tag-coll (flatten [tags])]


sanel:
Unless I'm mistaken, because of this [1] it can accept strings that will be converted to collection, so docstring and tests are correct. This is also confirmed by tests [2] and [3].

I'm not sure how you got that `(tagged-any "foo" prn)` emits each character, but after quick testing on 0.3.6 this is how it behaves:

```clojure
; riemann.config
(let [index (index)]
  (streams
   (tagged-any "test" prn)))
```

```sh
# shell
riemann-client send -S "test/test" -t test -T 127.0.0.1

# riemann.log
riemann.codec.Event{:host nil, :service "test/test", :state nil, :description nil, :metric nil, :tags ["test"], :time 1.652525815269E9, :ttl nil, :x-client "riemann-c-client"}
```

[1] https://github.com/riemann/riemann/blob/main/src/riemann/streams.clj#L1301
[2] https://github.com/riemann/riemann/blob/main/test/riemann/streams_test.clj#L263
[3] https://github.com/riemann/riemann/blob/main/test/riemann/streams_test.clj#L289

jarpy:
You are not mistaken, I indeed am! :)

I did all my testing with the predicate functions and somehow missed the extended behavior of the stream functions (even though, when I look at my editor, me last test config was using a stream function).

What I might do is come back with a patch that allows the predicate functions to also take a string, because the difference in behavior between them and stream functions surprises people.

Thanks very much for the review and detailed comments.

jarpy:
Raised #1015 to correct the little copy/paste error.

Closing this one.

jarpy:
> What I might do is come back with a patch that allows the predicate functions to also take a string[..]

Hmmm... never mind. I see from #1011 that the predicate functions are performance-critical, since, naturally, they are called for every event. So, yeah, I think the docs just a need a little bit of love to help new users understand the differences.

PR #1015, opened 2022-05-14
Fix cut-n-paste error in tagged-any docstring

jarpy:
This looks like a simple copy/paste error in a docstring. I think it's worth fixing because, to a new reader, it looks like it's comparing syntactic differences between `tagged-all` and `tagged-any`.


jamtur01:
Thanks!

PR #1026, opened 2022-08-04
add prefix to graphite tags

skob:
if graphite tags are `service` or something like, it's just silently disappears


jamtur01:
Can you give me some more information on what you're seeing here?

skob:
if event from graphite come with tags from reserved keywords (like `my.series;service=value1;host=value2 ` etc), then event will become {:host "value2", :service "value1"}

PR #1029, opened 2022-09-22
Fix expectation/test order in the test section

smortex:
= expect the expected result as first argument and the actual code as
second argument, otherwise the test output is inconsistent:

```
FAIL in (index-test) (riemann.config:43)
expected: {}
  actual: {:index [{:service "test", :time 1, :ttl 3}]}
    diff: + {:index [{:service "test", :time 1, :ttl 3}]}
```

Switch argument position so that test output is correct:

```
FAIL in (index-test) (riemann.config:43)
expected: {:index [{:service "test", :time 1, :ttl 3}]}
  actual: {}
    diff: - {:index [{:service "test", :time 1, :ttl 3}]}
```

### DIFF
--- a/howto.html
+++ b/howto.html
@@ -903,11 +903,12 @@ java -Dlogback.configurationFile=file:///path/to/logback.xml \
 {% highlight clj %}
 (tests
   (deftest index-test
-    (is (= (inject! [{:service "test"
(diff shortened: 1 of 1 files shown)


jamtur01:
Thanks!

PR #1033, opened 2022-12-15
Increase performance of Elasticsearch bulk forwarder.

cresh:
We see drastic performance problems when using the Elasticsearch bulk forwarder.  The ineffecient `reduce` in `gen-request-bulk-body` causes the problem because it uses `str` and thus copies the growing string for every element. Using `concat` and `string/join` instead shows a gain in performance of more than 90% in our benchmarks.


jamtur01 (approved):
LGTM

PR #1040, opened 2023-12-06
Also run the test suite on PR

smortex:
We have CI running when code is pushed to the riemann repository.  For
external contribution, this only happen when a PR is merged, but
detecting issues earlier is advisable.

Instead of running CI when code is pushed in any branch of the repo, run
CI when the main branch is updated, and when a PR targeting the main
branch is updated.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,12 @@
 name: Riemann testing
 
-on: [push]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
 
 jobs:
   test:


jamtur01:
Thanks @smortex - there was a reason for this in the dim dark past, but I cannot remember what it is so ... yolo. :)

smortex:
I guess it can be used by cryptobro to mine some cryptocrap and be quickly banned from GitHub :shrug: …

I have experienced some org that have tuned what happen when new contributors send their first PR: https://github.com/orgs/community/discussions/14334

I see there are option in *Settings* > *Actions* > *General* > *Fork pull request workflows from outside collaborators*, I have not changed the default but maybe you can check these?

PR #1046, opened 2024-10-31
Remove text-justify: distribute.

atw-latacora:
Per <https://developer.mozilla.org/en-US/docs/Web/CSS/text-justify>, "distribute" is deprecated, and has the same effect as inter-character.

Firefox 131.0.3 seems to respect text-justify: distribute. Chromium 129.0.6668.100 seems to ignore text-justify altogether; see <https://caniuse.com/css-text-justify>.

By omitting text-justify: distribute, Firefox and Chrom* ought to use their default justification, which adjusts spacing between words rather than between characters.

---

Subjectively, I think this looks better. e.g. in the second set of screenshots, in the "before", the line of text "bottlenecks in production. For everyone fed up with traditional" stands out oddly.

Screenshots:

| Before | After |
|--------|--------|
| Screenshot 2024-10-31 at 10 23 21 | Screenshot 2024-10-31 at 10 23 43 |
| Screenshot 2024-10-31 at 10 22 30 | Screenshot 2024-10-31 at 10 22 49 |

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -38,7 +38,6 @@ body {
   font-size: 17px;
   text-shadow: 0px 1px 3px rgba(0,0,0,0.3);
   text-align: justify;
-  text-justify: distribute;
   line-height: 26px;
  
   background: #cf004e; /* Old browsers */


smortex (approved):
Ah!  That probably explain why I feel something is odd and that the text is a bit hard to read sometimes.  The change is subtle, but provide more readability IMHO.

jamtur01:
Thanks!

PR #1049, opened 2025-01-07
Update packaging to produce EL9 packages too

faxm0dem:


smortex (approved):
That's probably fine. Luckily I don't have rpm-based systems anymore so I can't test it :-D

I guess we can remove RHEL 6 as it has reached EOL, and at some point adding RHEL 10 will make sense as it seems to be in beta ATM. I guess this is something for another PR, and maybe some refactoring could help avoid to have those long similar fatrpmX files.

faxm0dem:
We can drop EL7 too, it's already EOL

jamtur01:
Thanks!

PR #1050, opened 2025-04-03
Fix: Closing alerts on Opsgenie fails if alias contains special characters

holograph:
An Opsgenie alert can include an arbitrary custom alias (dedup key). The Opsgenie/JSM REST [API for closing alerts](https://docs.opsgenie.com/docs/alert-api#close-alert) supports providing an alias via a path parameter, requiring URL-encoding the alias, otherwise closing alerts can silently fail.


Graham-Russell-Forter (approved):

jamtur01:
Looks good! Thanks for this @holograph.

PR #1053, opened 2025-07-30
Switch to systemd on Debian

smortex:
We previously used a legacy init script to start / stop the riemann daemon on Debian.  The RPM packages use systemd, and it's time to embrace systemd on Debian too.


A-BenMao:
So when will this fetch request be integrated into the main storyline? Is there a timeline for it?

jamtur01:
Thanks @smortex

